PR #41, opened 2015-02-18
Date filtering

vforgione:
Added handling for `before` and `after` params.

* explicit range: `?before=2015-01-01&after=2014-12-01&note=only published in dec 2014`
* before a date: `?before=2015-01-01&note=everything published before 2015`
* after a date: `?after=2014-12-01&note=everything published after dec 2014`

I also linted the effected files.

### DIFF
--- a/bulbs/content/views.py
+++ b/bulbs/content/views.py
@@ -10,7 +10,6 @@ from django.views.generic import ListView, DetailView, View
 
 from bulbs.content.models import Content, ObfuscatedUrlInfo
 
-
 logger = logging.getLogger(__name__)
 
 
--- a/bulbs/content/views.py
+++ b/bulbs/content/views.py
@@ -1,7 +1,5 @@
 import logging
 
-from bulbs.content.models import Content, ObfuscatedUrlInfo
-
 from django.conf import settings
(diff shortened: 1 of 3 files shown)


bulbs/content/views.py
```diff
@@ -21,65 +21,79 @@ class ContentListView(ListView):
...
+            search_kwargs["authors"] = self.request.GET.getlist("authors", [])
+        elif "author" in self.request.GET:
+            search_kwargs["authors"] = self.request.GET.getlist("author", [])
+        if "authors" in self.kwargs:
+            search_kwargs["authors"] = self.kwargs["authors"]
         if self.authors:
-            search_kwargs['authors'] = self.authors
+            search_kwargs["authors"] = self.authors
+        
+        if "before" in self.request.GET:
```
csinchok: What formats are we supporting here? Won't we have to parse this to a datetime?

vforgione:
@csinchok: made a new function in the models to parse out tons of date like objects. it should be able to handle datetime, date, iso strings and strings that look like time stamps. i added tests for these as well.

csinchok:
LGTM! :+1:

PR #46, opened 2015-03-23
added videohub integration

vforgione:


vforgione:
don't need this any more

PR #49, opened 2015-04-03
Percolator updates

csinchok:
@vforgione I made a couple updates to the percolator code. Can you give it a once-over when you get a chance?

  - I found a bug where when you saved a percolator the *second* time, you got a 409 "Object Already Exists", since we were using `es.create`, instead of `es.index`.
  - In order to fulfill our requirements around the priority os sponsored sections, it would be nice to have a way to be able to check if a special coverage section is sponsored, just using the percolator response, so I added some of that metadata to the percolator payload. This should save us some db queries.

@kand If you check out [`SpecialCoverageQueryTests`](https://github.com/theonion/django-bulbs/blob/da7fbff59f79fcbe9344807b86fc636bf5b4aff0/tests/special_coverage/test_spec_cov_es.py#L26), you can see an example of how to build a query from scratch, instead of using `BaseCustomSearchFilterTests`. This might help with those issues we were talking about earlier?


vforgione:
Sorry, I just got around to checking this now. Everything looks splendid! :+1:

PR #51, opened 2015-04-23
Special coverage cascade

benghaziboy:
@kand


kand:
@benghaziboy looks good, merge it in.

PR #58, opened 2015-06-23
Added management command to change zone_length of a pzone

kand:
**New**
- set_pzone_zone_length management command added that allows you to change the length of a pzone


kand:
@vforgione here's the moved and updated code if you can review.

vforgione:
:+1:

PR #67, opened 2015-08-19
Validating special coverage queries

benghaziboy:


benghaziboy:
@akos this should prevent that clickhole issue from happening again.

tests/special_coverage/test_spec_cov_camp.py
```diff
@@ -100,3 +100,31 @@ def test_query_string_resolves_to_dict_when_saving(self):
         )
 
         self.assertIsInstance(SpecialCoverage.objects.all()[0].query, dict)
+
+    def test_query_string_filters_null_values_when_saving(self):
+        """Andrew Kos comments on tests. I however, do not."""
```
benghaziboy: @kand

kand: @benghaziboy are you going to turn that into a test or what? Your clarification is clearly not executable.

benghaziboy: I think it is important that I express myself in ways other than code.

benghaziboy: So does my hateful therapist.

PR #69, opened 2015-09-02
Make Section ES Indexable

mparent61:
@collin @benghaziboy 

This will not parse/index the somewhat-arbitrary "query" JSON content -- would we ever need this?


collin:
I'm building my env with django-bulbs pointed at branch out now to confirm it does what I need.

benghaziboy:
should be good. Might want to add a test to confirm mapping, but we can go through that later.
:+1:

vforgione:
:+1:

collin:
@mparent61 @benghaziboy @vforgione Mapped the slug. Made it possible to facet on sections.

bulbs/sections/models.py
```diff
@@ -26,6 +28,9 @@ class Section(models.Model):
 
     class Mapping:
         name = field.String(analyzer="autocomplete", fields={"raw": field.String(index="not_analyzed")})
+        slug = field.String(index="not_analyzed")
```
collin: slug map

PR #70, opened 2015-09-09
Validate pzones

benghaziboy:
@mparent61 @vforgione @MichaelButkovic 

I saw from some Starwipe stuff that we expect PZone data to be formatted as such

`PZone.data = [{'id': 1}, {'id': 2}]`

however were doing nothing to validate that, so here's this


MichaelButkovic:
LGTM

vforgione:
:+1:

PR #72, opened 2015-09-28
Add special coverage config

mparent61:
Customizable per-site (property) 

Will initially be used to disable Onion right rail video

@benghaziboy


benghaziboy:
Do we want to add this to the serializer?

mparent61:
Good catch, I always forget serializers. 

I also fixed the default to be an empty dict instead of "False"

benghaziboy:
pending tests,
LGTM

PR #75, opened 2015-10-02
Fix section percolator

benghaziboy:
@mparent61 @MichaelButkovic


MichaelButkovic:
:+1:

mparent61:
:+1:

benghaziboy:
@mparent61 @MichaelButkovic

bulbs/utils/percolator.py
```diff
@@ -0,0 +1,70 @@
...
+from elasticsearch_dsl.connections import connections
+
+from bulbs.content.models import Content
+from bulbs.sections.models import Section
+from bulbs.special_coverage.models import SpecialCoverage
+
+
+PERCOLATOR_PATTERNS = {
+    'section.(\w+)': Section,
+    'specialcoverage.(\w+)': SpecialCoverage
```
mparent61: AFAIK, should use r'' string prefix to make sure "\w" is correctly interpreted

bulbs/utils/percolator.py
```diff
@@ -0,0 +1,70 @@
...
+}
+
+
+def get_query_id_list(es, index, doc_type):
+    """Searches an elasticsearch index for all results of a given doc_type"""
+
+    results = es.search(index=index, doc_type=doc_type)
+    results_hits = results.get('hits', None)
+
+    if results_hits is None:
```
mparent61: probably better to do `if not result_hits:`, in case ES results ever return empty dict or list

bulbs/utils/percolator.py
```diff
@@ -0,0 +1,70 @@
...
+        raise KeyError('No queries matched in the index {}'.format(index))
+
+    hits = results_hits.get('hits', None)
+    if hits is None:
```
mparent61: same

bulbs/utils/percolator.py
```diff
@@ -0,0 +1,70 @@
...
+        cls = get_percolator_class(es_id)
+        if cls and not class_percolator_is_valid(cls, es_id):
+            bad_queries.append(es_id)
+    return bad_queries
+
+
+def clean_deprecated_class_percolators():
+    index = Content.search_objects.mapping.index
+    doc_type = '.percolator'
+    es = connections.get_connection('default')
```
mparent61: To avoid duplicate config with get_deprecated_config (and maintenance overhead), simpler to just call get_deprecated_queries with a callback, 
one callback would print() (for "--check" mode) the other would do the delete.
This would also simplify the command logic, it would call the same get_deprecated_queries def in both cases, just with different callback.

mparent61:
Closing per @benghaziboy

PR #76, opened 2015-10-06
special coverage rss feed

MichaelButkovic:
@spra85 @benghaziboy @mparent61

### DIFF
--- a/bulbs/feeds/views.py
+++ b/bulbs/feeds/views.py
@@ -45,21 +45,9 @@ class RSSView(ContentListView):
         return RequestContext(self.request, context)
 
 
-class SpecialCoverageRSSView(ContentListView):
+class SpecialCoverageRSSView(RSSView):
     """Really simply, this syndicates Content."""
-    template_name = "feeds/rss.xml"
-    paginate_by = 20
     feed_title = "Special Coverage RSS Feed"
-    utm_params = "utm_medium=RSS&amp;utm_campaign=feeds"
-
-    def get_template_names(self):
-        return ["feeds/rss.xml", "feeds/_rss.xml"]
-
-    @cache_control(max_age=600)
-    def get(self, request, *args, **kwargs):
-        response = super(SpecialCoverageRSSView, self).get(request, *args, **kwargs)
-        response["Content-Type"] = "application/rss+xml"
-        return response
 
     def get_queryset(self):
         sc_id = self.request.GET.get("special_coverage_id")
@@ -72,21 +60,4 @@ class SpecialCoverageRSSView(ContentListView):
         else:
             sc_ids = []
 
-        return super(SpecialCoverageRSSView, self).get_queryset().filter(Ids(values=sc_ids))[:self.paginate_by]
-
-    def get_context_data(self, *args, **kwargs):
-        context = super(SpecialCoverageRSSView, self).get_context_data(*args, **kwargs)
-        context["full"] = (self.request.GET.get("full", "false").lower() == "true")
-        context["images"] = (self.request.GET.get("images", "false").lower() == "true")
-        context["build_date"] = now()
-        context["title"] = self.feed_title
-        context["feed_url"] = self.request.build_absolute_uri()
-        context["search_url"] = self.request.build_absolute_uri(
-            u"/search?%s" % self.request.META["QUERY_STRING"])
-
-        # OK, so this is kinda brutal. Stay with me here.
-        for content in context["page_obj"].object_list:
-            feed_path = content.get_absolute_url() + "?" + self.utm_params
-            content.feed_url = self.request.build_absolute_uri(feed_path)
-
-        return RequestContext(self.request, context)
+        return ContentListView.get_queryset(self).filter(Ids(values=sc_ids))[:self.paginate_by]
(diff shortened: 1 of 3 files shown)


bulbs/feeds/views.py
```diff
@@ -40,3 +43,52 @@ def get_context_data(self, *args, **kwargs):
...
+        if sc_id and sc_slug:
+            sc_ids = SpecialCoverage.objects.get(id=sc_id, slug=sc_slug).query["included_ids"]
+        elif sc_id:
+            sc_ids = SpecialCoverage.objects.get(id=sc_id).query["included_ids"]
+        elif sc_slug:
+            sc_ids = SpecialCoverage.objects.get(slug=sc_slug).query["included_ids"]
+        else:
+            sc_ids = []
+
+        return super(SpecialCoverageRSSView, self).get_queryset().filter(Ids(values=sc_ids))[:self.paginate_by]
```
benghaziboy: custd

MichaelButkovic: :+1:

benghaziboy:
will this need the same filtering of external links as I have been seeing around re: RSS feeds

MichaelButkovic:
not as far as I know, but i'll defer to @spra85 on that

spra85:
I wouldn't think that external link items can be set up in a special coverage list, but maybe worth confirming at least.

bulbs/feeds/views.py
```diff
@@ -40,3 +43,52 @@ def get_context_data(self, *args, **kwargs):
...
+        return ["feeds/rss.xml", "feeds/_rss.xml"]
+
+    @cache_control(max_age=600)
+    def get(self, request, *args, **kwargs):
+        response = super(SpecialCoverageRSSView, self).get(request, *args, **kwargs)
+        response["Content-Type"] = "application/rss+xml"
+        return response
+
+    def get_queryset(self):
+        sc_id = self.request.GET.get("id")
```
spra85: Can we prefix these parameters with something a little more descriptive?  Maybe `special_coverage_id` / `special_coverage_slug`

bulbs/feeds/views.py
```diff
@@ -40,3 +43,52 @@ def get_context_data(self, *args, **kwargs):
...
+        sc_slug = self.request.GET.get("slug")
+
+        if sc_id and sc_slug:
+            sc_ids = SpecialCoverage.objects.get(id=sc_id, slug=sc_slug).query["included_ids"]
```
spra85: I don't think we'd ever query on both..maybe just need either id or slug

bulbs/feeds/views.py
MichaelButkovic: :+1: updated

bulbs/feeds/views.py
MichaelButkovic: sounds good to me. i was just going off the story description.

bulbs/feeds/views.py
```diff
@@ -40,3 +43,52 @@ def get_context_data(self, *args, **kwargs):
             content.feed_url = self.request.build_absolute_uri(feed_path)
 
         return RequestContext(self.request, context)
+
+
+class SpecialCoverageRSSView(ContentListView):
```
mparent61: If you inherit from RSSView you'd avoid a lot of duplicated code.

MichaelButkovic: ah, yeah, good point. i'll update

spra85:
:+1:

bulbs/feeds/views.py
MichaelButkovic: Updated!

bulbs/feeds/views.py
```diff
@@ -40,3 +43,21 @@ def get_context_data(self, *args, **kwargs):
...
+        sc_slug = self.request.GET.get("special_coverage_slug")
+
+        if sc_id:
+            sc_ids = SpecialCoverage.objects.get(id=sc_id).query["included_ids"]
+        elif sc_slug:
+            sc_ids = SpecialCoverage.objects.get(slug=sc_slug).query["included_ids"]
+        else:
+            sc_ids = []
+
+        return ContentListView.get_queryset(self).filter(Ids(values=sc_ids))[:self.paginate_by]
```
mparent61: wrong base class :)

MichaelButkovic: oops

bulbs/feeds/views.py
```diff
@@ -40,3 +43,21 @@ def get_context_data(self, *args, **kwargs):
...
+    def get_queryset(self):
+        sc_id = self.request.GET.get("special_coverage_id")
+        sc_slug = self.request.GET.get("special_coverage_slug")
+
+        if sc_id:
+            sc_ids = SpecialCoverage.objects.get(id=sc_id).query["included_ids"]
+        elif sc_slug:
+            sc_ids = SpecialCoverage.objects.get(slug=sc_slug).query["included_ids"]
+        else:
+            sc_ids = []
```
mparent61: Not 404ing on missing/invalid slug/id is intentional, right? To ensure we always provide Taboola with SOME content?

MichaelButkovic: that was the idea, yeah. if you think a 404 is better though, i can change it.

mparent61: @spra85 ?

spra85: I think we actually want to deliver an empty RSS in this scenario.  If they provide a slug or ID that doesn't map to anything, it should just be empty

mparent61:
:+1:

PR #83, opened 2015-11-10
added published field to special coverage model

collin:
@mparent61 @MichaelButkovic @benghaziboy 

For 2015 Our Annual Year we want to have a published date for special coverage sections so we can have cool query logic to drive the daily roll-out of the best-of-<month> special coverage sections.

I also added a `last_modified` field because I'm a habitual timestamper.


MichaelButkovic:
I also have a PR with a migration for Special Coverage, so we'll have to coordinate.

MichaelButkovic:
LGTM, otherwise. May want to add a quick test to make sure everything works as expected.

collin:
Will do.

Also I'll let you go first and re-write with a dependency on yours. The SW/Duracell SC is happening first.

mparent61:
:+1:

benghaziboy:
@collin I think we can delete this PR since I have added start & end logic to special coverage

PR #84, opened 2015-11-10
Retrieve content via absolute URL lookup

mparent61:
@MichaelButkovic  - This works, but I'm a REST novice, so please advise. 

- Using the list method to return a single object (which seems weird) 
- Passing the URL by query param seems cleaner than including it as part of the url path, since it will contain backslashes. 

A separate endpoint seemed like a cleaner approach, after exploring adding a filter to existing Content endpoint.

### DIFF
--- a/bulbs/api/views.py
+++ b/bulbs/api/views.py
@@ -2,8 +2,10 @@
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
+from django.core.urlresolvers import resolve, Resolver404
 from django.db.models.loading import get_models
 from django.http import Http404
+from django.shortcuts import get_object_or_404
 from django.utils import timezone
 from django.utils.dateparse import parse_datetime
 from djes.apps import indexable_registry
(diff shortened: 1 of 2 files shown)


mparent61:
@clowe why not...

bulbs/api/views.py
```diff
@@ -439,6 +441,21 @@ def list(self, request):
         return Response(dict(results=results))
 
 
+class ContentResolveViewSet(viewsets.ReadOnlyModelViewSet):
+    """Retrieve content corresponding to absolute URL"""
+    model = Content
+
+    def list(self, request):
+        if "url" in get_query_params(self.request):
```
MichaelButkovic: May want to throw a 400 or something if `url` isn't found.

MichaelButkovic:
LGTM

bulbs/api/views.py
mparent61: good idea

mparent61:
@spra85 look OK?

tests/api/test_content_api.py
```diff
@@ -581,3 +581,27 @@ def test_search(self):
...
+
+
+class TestContentResolveAPI(BaseAPITestCase):
+
+    def test_resolve_url(self):
+        make_content(id=123)
+        r = self.api_client.get(reverse("content-resolve-list"),
+                                dict(url="/r/123"), format="json")
+        self.assertEqual(r.status_code, 200)
+        self.assertEqual(123, r.data['id'])
```
spra85: Should we add one more test that's not just the `/r/<id>` shorthand, e.g. http://www.theonion.com/article/man-arriving-late-forced-use-excuse-he-was-saving--51824

spra85: Also, maybe one just to be safe with a bunch of UTM or other parameters tacked onto the end.

spra85:
Looks good, just the couple of thoughts on a few add'l tests to make sure we're buttoned up

tests/api/test_content_api.py
mparent61: Yeah i'll throw the params on.
The longer URLs are property-specific and not defined in django-bulbs, but might make sense to mock one up.

PR #88, opened 2015-12-07
Rss views

benghaziboy:
bad queryset pagination.

don't need this explicit return. Seems like it was a bandage that is no longer needed.

@mparent61 @MichaelButkovic


MichaelButkovic:
:+1:

tests/test_rss_feeds.py
```diff
@@ -28,6 +27,29 @@ def test_rss_feed(self):
...
+                title='Content #{}'.format(i),
+                published=timezone.now() - timezone.timedelta(days=i)
+            )
+        TestContentObj.search_objects.refresh()
+
+        client = Client()
+        resp = client.get(endpoint)
+        self.assertEqual(resp.status_code, 200)
+        page1_content_list = resp.context_data.get('content_list')
+        self.assertEqual(len(page1_content_list), 20)
```
mparent61: More thorough to check expected object ID list (vs length + "not in") checks. Could stagger publish times to get fixed ordering.

PR #96, opened 2016-01-11
bulbs-poll

collin:
@mparent @benghaziboy @MichaelButkovic @kand

Added notes for running django-bulbs against the onion-services elasticsearch.

Is this clear enough?

### TODO
- [x] move things like `'Intentionally blank'` into constants
- [x] setup linter
       Ran linter locally on new `poll/` files and tests
- [x] hoist `SodaheadAPIError` out of `Poll` namespace
- [x] hoist `sodahead_id` generation into function
- [x] use `if not self.sodahead_id`
- [x] favor `==` over `is` for equality checking (vs identity checking)
  - [x] some string comparisons
  - [x] some integer comparisons (http status codes)
- [x] write comment explaining rational behind sodahead integration
- [ ] re-open bengazi investigation to get to the truth
- [x] test serializer against explicit string values
- [ ] class level `@mock_vault(SECRETS)`
      @mparent61 some python 2/3 issues around copying the mock_vault implementation over
         I'm going to hold off on that for now.


example/settings.py
```diff
@@ -82,9 +82,15 @@
 
 ES_DISABLED = False
 
-ES_URLS = ['http://localhost:9200']
+ES_URLS = ['http://elasticsearch.local:9200']
```
benghaziboy: would probably prefer to keep this general and let the host's port forwarding of 9200 handle this

collin: Agree, but I wasn't able to get it forwarded on localhost. Do you know what docker-compose setting to get it forwarded properly?

kand: Not 100% sure if this is what we're looking for here, but you can set environment variables in `docker-compose` and then access that in `settings.py`.
- [Setting environment variables in `docker-compose`](https://github.com/theonion/starwipe-cms/blob/master/resources/docker/docker-compose.yml#L9)
- [Accessing environment variables in a settings file](https://github.com/theonion/starwipe/blob/ea59243679ea9c9dec7186c9654d92ad666d3612/starwipe/settings/testing.py#L6)

mparent61: Since django-bulbs doesn't use Docker, I also think we should keep this to use "localhost" by default. But I agree it would be nice to leverage the elasticsearch service running inside Docker.
Couple ideas: 
1) override default host with ENV variable
2) separate settings file with alternate host config
3) Is there a way to somehow forward port 9200 to the docker VM?

benghaziboy: We can forward dc elasticsearch to localhost:9200 and I believe should since that is how elasticsearch is run anyway, and like @mparent61 said, this isn't a docker project...

bulbs/poll/migrations/0002_answer.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('poll', '0001_initial'),
```
benghaziboy: Since these are part of the same initial commit, we should erase these migrations and create `0001_initial.py` again so that both the answer and poll migrations are there

bulbs/poll/models.py
```diff
@@ -0,0 +1,9 @@
+from django.db import models
+
+class Poll(models.Model):
```
benghaziboy: Is the Poll not going to be part of the general Content schema? Should we be able to search for it in the index?

benghaziboy: ```class Poll(Indexable)```

```class Answer(Indexable)```

**Comment by collin** [comment]
@benghaziboy @mparent61 This has changed a bit since the last request for comments. Can we get another look?

File: bulbs/poll/api.py [hunk 4]
```diff
@@ -0,0 +1,16 @@
+from rest_framework.routers import DefaultRouter
+
+from .viewsets import PollViewSet, AnswerViewSet
+
+router = DefaultRouter(trailing_slash=True)
+
+api_v1_router = DefaultRouter()
+api_v1_router.register(
```
benghaziboy: out of curiosity, do you have a pep8 linter?

bulbs/poll/models.py
```diff
@@ -0,0 +1,94 @@
...
+from rest_framework.exceptions import APIException
+
+from time import mktime
+import pytz
+import requests
+
+SODAHEAD_DATE_FORMAT = '%m/%d/%y %I:%M %p'
+
+class Poll(Content):
+    class SodaheadResponseError(APIException):
```
benghaziboy: Why does this class need to be nested in the model?

bulbs/poll/models.py
```diff
@@ -0,0 +1,94 @@
...
+            activation_date = self.published.astimezone(pytz.utc)
+            poll_payload['activationDate'] = activation_date.strftime(SODAHEAD_DATE_FORMAT)
+
+        if self.end_date:
+            end_date = self.end_date.astimezone(pytz.utc)
+            poll_payload['endDate'] = end_date.strftime(SODAHEAD_DATE_FORMAT)
+
+        for answer in self.answers.all():
+            if answer.answer_text is u'':
+                poll_payload[answer.sodahead_answer_id] = 'Intentionally blank'
```
benghaziboy: can we define these in constants?

bulbs/poll/models.py
```diff
@@ -0,0 +1,94 @@
...
+    answer_text = models.TextField(blank=True, default="")
+
+
+    def save(self, *args, **kwargs):
+        if self.sodahead_answer_id is "":
+            count = self.poll.last_answer_index
+            self.sodahead_answer_id = 'answer_%02d' % (count + 1)
+        super(Answer, self).save(*args, **kwargs)
+        self.poll.last_answer_index += 1
+        self.poll.save()
```
benghaziboy: Lotsa stuff going on here, let's make sure the response time doesn't balloon up. Potentially might wanna offload this celery

bulbs/poll/models.py
```diff
@@ -0,0 +1,94 @@
...
+    sodahead_answer_id = models.CharField(max_length=20, blank=True, default="")
+    answer_text = models.TextField(blank=True, default="")
+
+
+    def save(self, *args, **kwargs):
+        if self.sodahead_answer_id is "":
+            count = self.poll.last_answer_index
+            self.sodahead_answer_id = 'answer_%02d' % (count + 1)
+        super(Answer, self).save(*args, **kwargs)
+        self.poll.last_answer_index += 1
```
benghaziboy: why not let the id dictate this?

bulbs/poll/api.py
collin: No. What's a recommended way to run one? In-editor? Or could we install it to run as part of testing?

bulbs/poll/models.py
collin: Doesn't need to be, but we were taking a cue from errors like `ModelClass.DoesNotExist`.

bulbs/poll/models.py
collin: Yes.

bulbs/poll/api.py
benghaziboy: @mparent61 I think we use flake8, not sure how the plugin might work for vim, but :ahuh:

bulbs/poll/models.py
collin: This came up in analysis and it came out that we want our endpoint to return a failure code if sodahead fails. Extra time in the request was deemed acceptable.

bulbs/poll/models.py
benghaziboy: Usually, (qualifier) from my experiences, nested classes should be reserved for namespace conflicts as all it does is limit reusability.

bulbs/poll/models.py
collin: We don't want to do any resynchronization with the sodahead answers list in the event that we delete any of our answers so we need to have a predictable increment here.

Sodahead api demands answer keys look like this. :(

bulbs/poll/api.py
mparent61: Yeah we've been using flake8 with tunic/combine.

It's a good first-step linter that doest stuff we can all agree on. I recommend rolling this out to other projects.

mparent61: Combine also has an experimental precommit hook for Git that runs lint + tests. 

VIM has support for flake8 (I assume lots of editors do). Also created scripts/lint.

Flake8 is pretty popular and just is a wrapper around "pep8 and pyflakes" linters.

bulbs/poll/models.py
benghaziboy: fair enough. 

Can we pull the first 3 lines of the save function into its own

```def clean_sodehead_id``` method or something of the like.

File: bulbs/poll/models.py (hunk 8 above)
Comment (benghaziboy) [inline bulbs/poll/models.py]: cool. Figured it was legit, just wanted to make sure.

File: bulbs/poll/models.py (hunk 5 above)
Comment (mparent61) [inline bulbs/poll/models.py]: I would move it out, ModelClass.DoesNotExist pertains directly to ORM behavior tightly linked to model, whereas an API error falls outside ORM behavior.

File: bulbs/poll/models.py [hunk 9]
```diff
@@ -0,0 +1,94 @@
...
+        if not 'answer_01' in poll_payload:
+            poll_payload['answer_01'] = 'default answer 1'
+
+        if not 'answer_02' in poll_payload:
+            poll_payload['answer_02'] = 'default answer 2'
+
+        return poll_payload
+
+    def save(self, *args, **kwargs):
+        if self.sodahead_id is "":
```
mparent61: Shouldn't use "is" for comparing to empty string -- "is" is for comparing identity, "==" is for comparing value. 

Use "==", or better yet just "not" since it's ID is a CharField:      if not self.sodahead_id

collin: Thanks, will review for that.

bulbs/poll/models.py
mparent61: @collin I would add your explanation as a comment for future devs to understand the design decision.

benghaziboy:
I want to know what really happened at that compound

tests/poll/test_serializers.py
```diff
@@ -0,0 +1,72 @@
...
+        poll = Poll.objects.create(question_text=u'good text',
+                title=random_title())
+        answer1 = Answer.objects.create(poll=poll, answer_text=u'this is some text')
+        answer2 = Answer.objects.create(poll=poll, answer_text=u'forest path')
+        serializer = PollSerializer(poll)
+        answers_data = serializer.data['answers']
+        self.assertEqual(answers_data, [
+            {
+                'id': 1,
+                'answer_text': answer1.answer_text,
```
mparent61: In general I recommend explicitly putting the expected text string here, to make the expected out put more clear (and also to protect against rare chance where serializer might erroneously mutate object data.

tests/poll/test_models.py
```diff
@@ -0,0 +1,163 @@
...
+from .common import SECRETS
+
+vcr = make_vcr(__file__)  # Define vcr file path
+
+class PollTestCase(BaseIndexableTestCase):
+
+    """ Tests for the Poll model class """
+
+    @vcr.use_cassette()
+    @mock_vault(SECRETS)
```
mparent61: As future cleanup, I figured out how to have these mock_vaults apply at a class-level, so you only need to tag the entire class once.

https://github.com/theonion/combine/blob/master/combine/utils/tests.py#L76

bulbs/utils/vault.py
```diff
@@ -0,0 +1,28 @@
+# Hashicorp Vault client interface
```
kand: sweet :+1:

tests/poll/test_models.py
collin: Mike, you are a beautiful human being.

README.md
```diff
@@ -7,6 +7,8 @@ This project very much under active development, and is currently not as reusabl
 
 ## Configuring for Development
 
+Make sure an instance of elasticsearch is available at `http://elasticsearch.local:9200/`. If you are running theonion/onion-services you can add `elasticsearch.local` to your `/etc/hosts` line that points at that docker dev machine.
+
```
mparent61: I'm OK with this -- onion-services will be running for most everyone, and /etc/hosts is an easy workaround. In the end it streamlines dev setup -- no reason to have ES running natively when it's also running as a container.

bulbs/poll/models.py
```diff
@@ -0,0 +1,94 @@
...
+        response = requests.delete('https://onion.sodahead.com/api/polls/{}/?access_token={}'
+                .format(self.sodahead_id, vault.read('sodahead/token')))
+        if response.status_code is not 204:
+            raise Poll.SodaheadResponseError(response.text)
+        super(Poll, self).delete(*args, **kwargs)
+
+    def sync_sodahead(self):
+        response = requests.post('https://onion.sodahead.com/api/polls/{}/'
+                .format(self.sodahead_id), self.sodahead_payload())
+        if response.status_code is not 200:
```
mparent61: similar to empty strings comment, should use "!=" instead of "is not" for integer value comparisions

README.md
collin: @benghaziboy was not a fan.

We were able to us the `onion-services`  ES at `localhost:9200` by forwarding the port in VirtualBox.

I'm in favor of leaning on /etc/hosts but let's see if we can get consensus on this.

It is noteworthy that the stock ES in hombrew is at an incompatible version where onion-services ES is version locked.

bulbs/poll/models.py
collin: :+1: will the linter suggest these changes too?

bulbs/utils/test.py
```diff
@@ -25,13 +32,31 @@ def make_content(*args, **kwargs):
...
+def make_vcr(test_path, record_mode='once'):
+    base_dir = os.path.dirname(os.path.realpath(test_path))
+    test_name =  os.path.splitext(test_path)[0]
+    cassette_dir = os.path.join(base_dir, 'test_data/cassettes', test_name)
+
+    return vcr.VCR(
+            cassette_library_dir=cassette_dir,
+            ignore_hosts=['localhost'],
+            record_mode=record_mode,
+            filter_post_data_parameters=['access_token'],
```
mparent61: nice

bulbs/poll/models.py
mparent61: Good question. A heavyweight linter like "PyLint" would, not sure about
flake8.



On Mon, Jan 25, 2016 at 10:36 AM, collin <notifications@github.com> wrote:

> In bulbs/poll/models.py
> <https://github.com/theonion/django-bulbs/pull/96#discussion_r50718357>:
>
> > +        else:
> > +            self.sync_sodahead()
> > +
> > +        super(Poll, self).save(*args, **kwargs)
> > +
> > +    def delete(self, *args, **kwargs):
> > +        response = requests.delete('https://onion.sodahead.com/api/polls/{}/?access_token={}'
> > +                .format(self.sodahead_id, vault.read('sodahead/token')))
> > +        if response.status_code is not 204:
> > +            raise Poll.SodaheadResponseError(response.text)
> > +        super(Poll, self).delete(*args, **kwargs)
> > +
> > +    def sync_sodahead(self):
> > +        response = requests.post('https://onion.sodahead.com/api/polls/{}/'
> > +                .format(self.sodahead_id), self.sodahead_payload())
> > +        if response.status_code is not 200:
>
> [image: :+1:] will the linter suggest these changes too?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/theonion/django-bulbs/pull/96/files#r50718357>.
>



-- 
Mike Parent
Developer
Onion, Inc.

README.md
mparent61: Hmmm you'll have to show me how you forwarded that port -- that might cleanup our need to use /etc/hosts at all for our *.local dev sites too.

mparent61:
Spoke with Melissa, going to rollout v0.7.0 (breaking Special Coverage changes) that requires site updates applied today + Monday. This means this PR can't be merged + rolled out to sites until roughly Tue (2015-02-16). 

https://github.com/theonion/django-bulbs/pull/110

mparent61:
Also need to coordinate with @benghaziboy re: some potential issues with 0.6.6 (Contribution celery tasks) we're investigating.

Bottom line: Need to coordinate before this is pushed out to sites.

collin:
@mparent61 @benghaziboy @MichaelButkovic 

Alright, I've addressed most of the notes and some of the extra ones you had on the PR @MelizzaP  started for polls phase ||.

Any last notes?

bulbs/poll/models.py
```diff
@@ -0,0 +1,235 @@
...
+def Closed():  # noqa
+    end_date_params = {
+        "lte": timezone.now(),
+    }
+
+    return Range(end_date=end_date_params)
+
+
+class PollManager(ContentManager):
+    def search(self, **kwargs):
```
benghaziboy: niiiiiiiice

bulbs/poll/models.py
```diff
@@ -0,0 +1,235 @@
...
+        response = requests.get(SODAHEAD_POLL_ENDPOINT.format(self.sodahead_id))
+
+        if not response.ok:
+            raise SodaheadResponseError(response.text)
+        else:
+            return response.json()
+
+    def sodahead_payload(self):
+        payload = {
+            'access_token': vault.read('sodahead/token')['value'],
```
benghaziboy: ever the possibility of this not being here

bulbs/poll/models.py
```diff
@@ -0,0 +1,235 @@
...
+    poll = models.ForeignKey(
+        Poll,
+        on_delete=models.CASCADE,
+        related_name='answers'
+    )
+    sodahead_answer_id = models.CharField(max_length=20, blank=True, default="")
+    answer_text = models.TextField(blank=True, default="")
+
+    def save(self, *args, **kwargs):
+        # using transaction/select_for_update here because we don't want to
```
benghaziboy: @mparent61 do we care about docstring format for multiline?

bulbs/poll/models.py
collin: Yes, it is possible the vault could lose the value for some reason.

bulbs/poll/models.py
benghaziboy: or rather method/function level docstrings

benghaziboy: don't care to hold up the pr about it, figured i'd just bring it up

benghaziboy:
we need a way to silence all these fixtures in PRs, but other than that....
:+1:

collin:
@mparent61 :+1:  on coordination /cc @spra85

MichaelButkovic:
LGTM

MichaelButkovic:
actually no lgtm until `re-open bengazi investigation to get to the truth` is finished

bulbs/poll/models.py
mparent61: We should add caching + error handling inside vault client, as a followup PR

bulbs/poll/models.py
mparent61: I don't have a preference for docstrings vs multiline comments, and I just silence lint errors about missing docstrings. 
Any reason we should have a preference?

README.md
mparent61: Latest django-bulbs leans on ELASTICSEARCH_HOST environment variable, this feels much less invasive than modifying /etc/hosts.

bulbs/utils/test.py
```diff
@@ -25,13 +32,33 @@ def make_content(*args, **kwargs):
...
 
+def make_vcr(test_path, record_mode='once'):
+    base_dir = os.path.dirname(os.path.realpath(test_path))
+    test_name =  os.path.splitext(test_path)[0]
+    cassette_dir = os.path.join(base_dir, 'test_data/cassettes', test_name)
+
+    return vcr.VCR(
+            cassette_library_dir=cassette_dir,
+            ignore_hosts=[
+                'localhost',
```
mparent61: this won't ignore dockerized ES host

mparent61:
Spoke with @collin, going to review this THUR AM after merging with @MelizzaP's branch and related PR comments

bulbs/poll/models.py
```diff
@@ -0,0 +1,243 @@
...
+        super(Poll, self).save(*args, **kwargs)
+
+    def delete(self, *args, **kwargs):
+        response = requests.delete(
+            SODAHEAD_DELETE_POLL_ENDPOINT.format(
+                self.sodahead_id,
+                vault.read('sodahead/token')['value'],
+            )
+         )
+        if response.ok:
```
mparent61: You can drop these 2 lines, they don't do anything

bulbs/poll/models.py
```diff
@@ -0,0 +1,243 @@
...
+            raise SodaheadResponseFailure(response.text)
+        super(Poll, self).delete(*args, **kwargs)
+
+    def sync_sodahead(self):
+        response = requests.post(
+            SODAHEAD_POLL_ENDPOINT.format(self.sodahead_id),
+            self.sodahead_payload()
+        )
+
+        if response.ok:
```
mparent61: Same, don't need these 2 lines

mparent61: In this case, either drop this check, or do something like
```
if not response.ok:
     if response.status_code > 499:
        raise SodaheadResponseError(response.json())
    else:
        raise SodaheadResponseFailure(response.json())
```

bulbs/poll/models.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def save(self, *args, **kwargs):
+        if not self.sodahead_id:
+            response = requests.post(SODAHEAD_POLLS_ENDPOINT, self.sodahead_payload())
+
+            if response.ok:
+                self.sodahead_id = response.json()['poll']['id']
+            elif response.status_code > 499:
+                raise SodaheadResponseError(response.text)
+            elif response.status_code > 399:
```
mparent61: Make the last case just an "else" to ensure it handles "everything else"

bulbs/api/views.py
```diff
@@ -189,10 +189,10 @@ def trash(self, request, **kwargs):
         doc_type = content.__class__.search_objects.mapping.doc_type
 
         try:
-            Content.search_objects.client.delete(
-                index=index,
-                doc_type=doc_type,
-                id=content.id)
+            self.model.search_objects.client.delete(
+                    index=index,
```
mparent61: This is minor, but best not to introduce non-standard (PEP8) indentation to existing code throughout this file. All whitespace-only changes in this file should be reverted. Someday (soon!) we want a clean lint run on django-bulbs.

bulbs/api/views.py
```diff
@@ -450,8 +450,8 @@ def list(self, request):
                 results.append(dict(
                     name=name,
                     doctype=doctype
-                ))
-        results.sort(key=lambda x: x["name"])
+                    ))
+                results.sort(key=lambda x: x["name"])
```
mparent61: This is a significant whitespace change that should also be reverted, as it will cause sort to be called multiple times instead of just once after for-loop completes (unless this is intended?)

bulbs/poll/views.py
```diff
@@ -0,0 +1,33 @@
...
+class MergedPollDataView(DetailView):
+    model = Poll
+
+    def render_to_response(self, context, **response_kwargs):
+        serializer = PollPublicSerializer(self.object)
+        response = HttpResponse(
+            json.dumps(serializer.data),
+            content_type="application/json"
+        )
+        if "HTTP_ORIGIN" in self.request.META:
```
mparent61: Why is this necessary here? (Should document)

bulbs/utils/test.py
```diff
@@ -25,13 +32,31 @@ def make_content(*args, **kwargs):
...
+    return vcr.VCR(
+            cassette_library_dir=cassette_dir,
+            ignore_hosts=settings.ES_CONNECTIONS['default']['hosts'],
+            record_mode=record_mode,
+            filter_post_data_parameters=['access_token'],
+            filter_query_parameters=['access_token']
+    )
+
+def random_title():
+    return ''.join([random.choice(string.ascii_uppercase) for _ in range(10)])
```
mparent61: nice one-liner

bulbs/api/views.py
collin: I don't know what happened here.  I definitely don't recall going through this file and changing indentations ???

But it's definitely me from the blame.

Maybe this explains my fever code dreams ;)

Will look through the diff on this file and revert them.

bulbs/utils/test.py
collin: props go to some unknown stackoverflow user

mparent61:
One more nit: I'd move the VCR files into a vcr-themed subdirectory ("vcr", "cassettes", etc) to clearly separate them from similarly-named `test_*.py` test files. This just requires a simple tweak to the `make_vcr` helper function and will benefit future VCR users too.

mparent61:
Other than my comments, :+1:

bulbs/api/views.py
collin: Fixed this one you noted and made some intentional changes per flake8.

bulbs/poll/views.py
collin: Added docs (not showing up in-line here, maybe we broke the PR?)

collin:
I got the poll cassettes moved to `test/poll/test_data/cassettes`

MichaelButkovic:
:+1:

collin:
Alright @mparent61 I think we have this all covered.

What's our rollout pipeline looking like?

mparent61:
One Python 3 test is still failing: https://travis-ci.org/theonion/django-bulbs/jobs/113500999

PR #98, opened 2016-01-12
Special filters

benghaziboy:
So. The Client currently filters by 
`active=True` & `active=False`, which now reflects things in a window. I like that pattern actually...so I decided I'd let the backend handle this rather than update the client atm


benghaziboy:
@MichaelButkovic @mparent61

MichaelButkovic:
LGTM

bulbs/special_coverage/filters.py
```diff
@@ -0,0 +1,20 @@
...
+    Includes the default boolean_filter, but replaces the active and inactive states to represent
+    whether the current_date is within a SpecialCoverage's start_date & end_date range.
+    """
+    def filter_queryset(self, request, queryset, view):
+        queryset = super(SpecialCoverageFilter, self).filter_queryset(request, queryset, view)
+        query_params = get_query_params(request)
+        if "active" in query_params:
+            value = query_params.get("active").lower()
+            if value == "true":
+                queryset = queryset.filter(start_date__lte=today(), end_date__gte=today())
```
mparent61: I know this is being picky, but should save today() result to local variable and use for both params, to avoid (extremely) slim chance of today() returning two different values

mparent61:
I like re-using the active filter, simplifies client interaction

mparent61:
Otherwise :+1:

PR #102, opened 2016-01-19
fix dumb tasking

benghaziboy:
@mparent61 @MichaelButkovic


MichaelButkovic:
:+1:

bulbs/contributions/models.py
```diff
@@ -159,14 +160,17 @@ class ContributorRole(Indexable):
...
         super(ContributorRole, self).save(*args, **kwargs)
-        if self.payment_type == 1:
-            self.create_feature_type_rates()
-
-    def create_feature_type_rates(self):
-        for feature_type in FeatureType.objects.all():
-            rate = FeatureTypeRate.objects.filter(role=self, feature_type=feature_type)
-            if not rate.exists():
+        self.create_feature_type_rates(created)
+        update_role_rates.delay(self)
```
mparent61: I've always avoided sending objects over Celery (to avoid serialization errors) and instead just sent IDs. Have you had success passing objects to Celery w/ Django projects?

bulbs/contributions/views.py
```diff
@@ -36,10 +36,10 @@ class ContributorRoleViewSet(viewsets.ModelViewSet):
     serializer_class = ContributorRoleSerializer
 
     def get_queryset(self):
-        qs = ContributorRole.search_objects.all()
-        if self.request.QUERY_PARAMS.get('override', None) == 'true':
-            qs = qs.exclude(payment_type=3)
-        return qs
+        queryset = ContributorRole.objects.all()
+        if self.request.QUERY_PARAMS.get("override", None) == "true":
+            queryset = queryset.exclude(payment_type=3)
```
mparent61: I'd use an enum/constant instead of "3" or add a comment explaining what 3 is

bulbs/contributions/models.py
benghaziboy: This seemed to run fine in tests/local

mparent61: Web consensus is that Passing object is dangerous, PK is better. 
----
https://denibertovic.com/posts/celery-best-practices/
You shouldn't pass Database objects (for instance your User model) to a background task because the serialized object might contain stale data. What you want to do is feed the task the User id and have the task ask the database for a fresh User object.

mparent61: HAHA didn't mean to shout silly github

benghaziboy: was only a matter of time before @mparent61 lost it

benghaziboy:
Don't know why its ignoring me updating that to a constant, but it is. 

I had to move the task call to a signal because of circular imports

benghaziboy:
@mparent61

mparent61:
Lookin' good :+1:

PR #107, opened 2016-01-25
django-admin freelance editing

benghaziboy:
@MichaelButkovic @mparent61 

We are going to be designing a user edit cms page in the future, but for the time being, we need to edit their information somewhere.


MichaelButkovic:
:+1:

mparent61:
:+1: yeah i guess

PR #110, opened 2016-02-08
Special coverage percolator scoring

mparent61:
Adds sponsored special coverage precedence rules. Additionally:

- Moves special coverage percolator queries into bulbs
- Special Coverages ALWAYS percolated (fixes recent regression where in-active SC never added to percolator after start_date becomes active until it is saved). Must update property site logic to filter start dates when upgrading. 
- Lint

**This introduces breaking changes to Special Coverage percolator** that will require property sites to switch to using these new bulbs-based percolator methods (because all special coverages now percolated and must be filtered on percolator GET queries). A migration step will automatically trigger percolator re-saves. 

Extra effort required to make this all work under ES v1.4, which is missing various Percolator scoring and filtering features. 

[STD-2853](https://theonion.atlassian.net/secure/RapidBoard.jspa?rapidView=9&selectedIssue=STD-2853&quickFilter=55)


bulbs/special_coverage/models.py
```diff
@@ -104,11 +104,21 @@ def _save_percolator(self):
...
 
         # We'll need this data, to decide which special coverage section to use
+        q["sponsored"] = bool(self.campaign)
         if self.campaign:
-            q["sponsored"] = True
-            q["start_date"] = self.campaign.start_date
-            q["end_date"] = self.campaign.end_date
-
+            # ES v1.4 percolator doesn't support DateTime ranges, use integer epoch seconds instead
+            q["start_date"] = datetime_to_epoch_seconds(self.campaign.start_date)
```
mparent61: Or should these fields be renamed to "start_date_epoch" to make their integer type more explicit? Downside is more fields and need to add ES mapping

bulbs/utils/methods.py
```diff
@@ -52,3 +53,7 @@ def is_valid_digit(value):
     elif is_str(value):
         return value.isdigit()
     return False
+
+
+def datetime_to_epoch_seconds(value):
+    return time.mktime(value.timetuple())
```
benghaziboy: Test is failing in the event of a None value

mparent61: Thx, i gotta fix that!

bulbs/special_coverage/models.py
benghaziboy: can we make `start_date_epoch` a property and index that?

mparent61:
I need to sort out start_date/end_date field issues, otherwise rest of logic is in place.

bulbs/special_coverage/models.py
mparent61: sure

mparent61:
@benghaziboy @MichaelButkovic @spra85 

This is ready for review. 

Everything now works on ES v1.4+, and I Travis now tests against multiple ES versions (so we're ready for a future upgrade).

tests/special_coverage/test_spec_cov_camp.py
```diff
@@ -104,12 +104,24 @@ def test_query_string_resolves_to_dict_when_saving(self):
...
+                },
+            ],
+        }
         data = {
             "name": "special boy cammy",
             "query": {
                 "pinned_ids": [1, 2, 3, None],
                 "excluded_ids": [4, 5, 6, None],
-                "groups": ["fun guys", None],
+                "groups": [group, None],
```
mparent61: This was previously broken but hidden b/c SpecialCoverage._save_percolator() wasn't saving inactive special coverages to percolator

bulbs/special_coverage/models.py
```diff
@@ -83,10 +85,8 @@ def save(self, *args, **kwargs):
         super(SpecialCoverage, self).save(*args, **kwargs)
 
         if self.query and self.query != {}:
-            if self.is_active:
-                self._save_percolator()
-            else:
-                self._delete_percolator()
+            # Always save and require client to filter active date range
+            self._save_percolator()
```
mparent61: This fixes a regression w/ new is_active logic where a SpecialCoverage start_time would become active but it wouldn't be in percolator (b/c when it was last saved it wasn't active)

bulbs/special_coverage/models.py
```diff
@@ -138,8 +154,7 @@ def get_content(self, published=True):
     def is_active(self):
         now = today_as_utc_datetime()
         if self.start_date and self.end_date:
-            if self.start_date < now and self.end_date > now:
-                return True
+            return self.start_date <= now < self.end_date
```
mparent61: We've been following a "start <= now < end" elsewhere throughout codebase (vs "start < now < end")

MichaelButkovic:
:+1:

bulbs/content/models.py
```diff
@@ -337,8 +368,119 @@ def get_serializer_class(cls):
...
+        # sort via scoring_functions, and then manually filter out zero scores.
+        sponsored_filter = {
+            "query": {
+                "function_score": {
+                    "functions": [
+
+                        # Boost Recent Special Coverage
+                        # Base score is start time
+                        # Note: ES 1.4 sorting granularity is poor for times
+                        # within 1 hour of each other.
```
spra85: That scenario very rarely happens and when it does, it's generally special coverage with items that have no overlap

bulbs/content/models.py
```diff
@@ -337,8 +368,119 @@ def get_serializer_class(cls):
...
+                                "term": {
+                                    "sponsored": False,
+                                }
+                            },
+                            "weight": 0,
+                        },
+                        # Penalize Inactive
+                        {
+                            "filter": {
+                                "or": [
```
spra85: Maybe it's lower in the PR and I'll see it, but I wanted to make sure we're completely excluding inactive special coverage from consideration and not just penalizing

bulbs/content/models.py
```diff
@@ -337,8 +368,119 @@ def get_serializer_class(cls):
...
+        }
+
+        results = _percolate(index=self.mapping.index,
+                             doc_type=self.mapping.doc_type,
+                             content_id=self.id,
+                             body=sponsored_filter)
+
+        return [r["_id"] for r in results
+                # Zero score used to omit results via scoring function (ex: unsponsored)
+                if r['_score'] > 0]
```
spra85: We use unsponsored special coverage now - just want to make sure those aren't going to be missed with this if that score is 0 as well as the inactive ones

tests/content/test_content_percolator.py
```diff
@@ -0,0 +1,180 @@
...
+        self.assertEqual(self.content.percolate_sponsored_special_coverages(),
+                         ['specialcoverage.{}'.format(i) for i in expected_ids])
+
+    def test_empty(self):
+        self.check_special_coverages([])
+
+    def test_ignore_unsponsored(self):
+        for sponsored in [True, False]:
+            make_special_coverage(tag='white', start=-1, end=1, sponsored=sponsored)
+        self.check_special_coverages([1])
```
spra85: Per above, believe we need unsponsored

bulbs/content/models.py
mparent61: Good, I didn't think this would be a problem but could be a tricky "gotcha" if we ever have to debug an ordering problem.

mparent61:
@spra85  Updated per our discussion. Merged both percolator queries into a single method, that will optionally omit Unsponsored.

bulbs/content/models.py
mparent61: inactive special coverage is completely excluded, I updated the comments to note this

spra85:
:+1:

mparent61:
Updated query to sort Sponsored ahead of Unsponsored. This allows property reading lists to just call percolate_special_coverage() once,  instead of 2 separate calls   (#1-sponsored, #2-all)

https://github.com/theonion/django-bulbs/compare/973f593a29a286bab037312aa7d498be779f5cc7...cfedeeb1cd37a3cc4a3706ff9b1a0ad3dd8793e1

PR #111, opened 2016-02-09
initial implementation

benghaziboy:
@mparent61 

This will be imported into the respective reading_lists to add the appropriate query.

### DIFF
--- a/bulbs/special_coverage/search.py
+++ b/bulbs/special_coverage/search.py
@@ -1,0 +1,28 @@
+"""Custom search function to assist in customizing our special coverage searches."""
+
+
+def second_slot_query_generator(query1, query2):
+    """Returns the result of a different query at the 1st index of iteration.
+
+    :param query1: Primary search that will be the default result set.
+    :type  query1: Any iterable object; intended for djes.search.LazySearch and django.Querysets.
+    :param query2: Secondard search that will return at the 1st index of iteration.
+    :type  query2: Any iterable object; intended for djes.search.LazySearch and django.Querysets.
+    """
+    index = 0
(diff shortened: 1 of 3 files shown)


bulbs/special_coverage/search.py
```diff
@@ -0,0 +1,28 @@
...
+    :param query2: Secondard search that will return at the 1st index of iteration.
+    :type  query2: Any iterable object; intended for djes.search.LazySearch and django.Querysets.
+    """
+    index = 0
+    offset = 0
+    while True:
+        result = None
+        if index == 1:
+            try:
+                result = query2[0]
```
mparent61: Do queries support a next() iterface? Then you wouldn't need to track offset
`result = next(query2)`
(We may have been down this road before -- perhaps queries behave beset when indexed via []? )

benghaziboy: currently they do not.

mparent61:
Otherwise LGTM 
I think this is the most flexible approach. Only immediate downside is the extra code to check for IndexErrors and such -- maybe that could be streamlined by some wrapper if we continue this approach?

benghaziboy:
@mparent61 updated this to include the `next` calls from the most recent djes updates

mparent61:
Much cleaner with next().
As I think you mentioned in the djes update, will the next() call make repeated ES backend queries on subsequent calls against the same query? Or will caching ensure ES is only hit once?

benghaziboy:
@mparent is this good to go?

MichaelButkovic:
:+1:

benghaziboy:
blessings

PR #114, opened 2016-02-10
Signal config

benghaziboy:
@mparent61 @MichaelButkovic


example/settings.py
```diff
@@ -36,12 +36,12 @@
     "bulbs.redirects",
     "bulbs.cms_notifications",
     "bulbs.content",
-    "bulbs.contributions",
     "bulbs.promotion",
     "bulbs.special_coverage",
     "bulbs.sections",
     # local testing apps
     "example.testcontent",
+    "bulbs.contributions",
```
mparent61: why the move here, some sort of import ordering dependency?

bulbs/content/models.py
```diff
@@ -326,7 +327,9 @@ def save(self, *args, **kwargs):
             if kwargs is None:
                 kwargs = {}
             kwargs["index"] = False
-        return super(Content, self).save(*args, **kwargs)
+        content = super(Content, self).save(*args, **kwargs)
+        index_content_contributions.delay(self.id)
```
mparent61: So this isn't in a save signal handler b/c of the child signal thing?

benghaziboy: no. That fixed the problem locally. However, the signal is not being recognized on test. After debugging I decided to move it into the save function because this needs to be out the door and this is seemingly functional.

example/settings.py
benghaziboy: This was part of debugging. Nothing came of it.

benghaziboy: removing

mparent61:
:+1: 
Seems much cleaner than the child_class stuff.

benghaziboy:
Yeah I generally prefer avoiding signal logic anyway.

MichaelButkovic:
LGTM

mparent61:
Re :+1:

PR #119, opened 2016-02-16
Fix special coverage migration save

mparent61:
v0.7.3  Fix special coverage migration to grab model directly so that custom SpecialCoverage.save() method is called and triggers _save_percolator.

A model obtained from apps.get_model() calls the generic save(), and *does not* call a custom save() method like SpecialCoverage.save(). This surprised me, take heed when using app.get_models() in your migration scripts. 

https://realpython.com/blog/python/data-migrations/#comment-1843026722


mparent61:
@benghaziboy @MichaelButkovic

MichaelButkovic:
huh, very weird

MichaelButkovic:
LGTM

bulbs/special_coverage/migrations/0009_auto_20160208_1540.py
```diff
@@ -1,13 +1,14 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
-from django.db import migrations, models
+from django.db import migrations
+
+from bulbs.special_coverage.models import SpecialCoverage
```
benghaziboy: Have you re-run this locally? 

Migrations can have dependency issues if not explicit, could be wrong here.

benghaziboy:
provided this has run,

LGTM

mparent61:
Yeah runs locally.
I think the dependency issues involve cross-app dependencies. Also by not using get_model() you can run into trouble if your migration also modifies schema.

PR #121, opened 2016-02-16
Celery is now a regular (non-dev) requirement

mparent61:
Also cleanup unnecessary (and broken) task import error handler. This is a relic of a lost, magical time we'll never see again.

@benghaziboy as discussed (and updated DJES requirements to avoid breaking tests)

 @MichaelButkovic join the celery party

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,18 @@
 
 ## Development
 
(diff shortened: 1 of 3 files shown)


MichaelButkovic:
:tada: LGTM

benghaziboy:
:+1:

PR #124, opened 2016-02-18
Signal config

benghaziboy:
@mparent61 @MichaelButkovic 

Testing this out locally so there might be some changes coming, but this is the first go at a wrapper that runs a collective query for special coverage objects.


benghaziboy:
@mparent61 @MichaelButkovic

MichaelButkovic:
:+1:

bulbs/special_coverage/search.py
```diff
@@ -1,27 +1,110 @@
...
+        Add a given queryset to the iterator with custom logic for iteration.
+
+        :param queryset: List of objects included in the reading list.
+        :param validator: Custom logic to determine a queryset's position in a reading_list.
+            Validators must accept an index as an argument and return a truthy value.
+        :param default: Sets the given queryset as the primary queryset when no validator applies.
+        """
+        if default or self.default_queryset is None:
+            self.default_queryset = queryset
+        if validator:
```
mparent61: Should provide error/feedback if "no validator and not default", else it will silently ignore the queryset

bulbs/special_coverage/search.py
```diff
@@ -1,27 +1,110 @@
...
+        self.default_queryset = None
+        self.querysets = {}
+        self.index = 0
+
+    def __iter__(self):
+        return self
+
+    def next(self):
+        for validator, queryset in self.querysets.items():
+            if validator(self.index):
```
mparent61: I'd use an ordered dictionary for self.querysets to have predictable behavior when 2+ queryset validators return TRUE for same index

bulbs/special_coverage/utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+def get_sponsored_special_coverages():
+    """:returns: Django query for all active special coverages with active campaigns."""
+    now = timezone.now()
+    return SpecialCoverage.objects.filter(
+        campaign__isnull=False,
+        campaign__start_date__lte=now,
+        campaign__end_date__gte=now,
+        start_date__lte=now,
+        end_date__gte=now
```
mparent61: We should standardize on our "greater/less than" and "greater/less"

I think it's most intuitive to have "start" be inclusive and "end" be exclusive -- so  `start_date <= now < end`, which makes it easy to have adjacent ranges that don't overlap.

bulbs/content/custom_search.py
```diff
@@ -118,6 +118,43 @@ def filter_from_query(query, id_field="id", field_map={}):
...
+                    condition_filter &= Nested(path=path, filter=Term(**{field_name: value}))
+                else:
+                    condition_filter &= Term(**{field_name: value})
+        elif operation == "any":
+            if "." in field_name:
+                path = field_name.split(".")[0]
+                condition_filter &= Nested(path=path, filter=Terms(**{field_name: values}))
+            else:
+                condition_filter &= Terms(**{field_name: values})
+        elif operation == "none":
```
mparent61: raise exception on invalid operation?

bulbs/special_coverage/search.py
```diff
@@ -1,27 +1,110 @@
...
-        yield result
-        index += 1
+from bulbs.content.custom_search import get_condition_filter
+from bulbs.content.models import Content
+
+
+class SearchParty(object):
+    """Wrapper that searches for content from a group of SpecialCoverage objects."""
+
+    def __init__(self, special_coverages, *args, **kwargs):
```
mparent61: Might as well drop *args + **kwargs since not using them, else they could mask unused params passed by caller

bulbs/special_coverage/search.py
```diff
@@ -1,27 +1,110 @@
...
+                "pinned_ids": [],
+                "groups": [],
+            })
+            for special_coverage in self._special_coverages:
+                # Access query at dict level.
+                query = getattr(special_coverage, "query", {})
+                if "query" in query:
+                    query = query.get("query")
+
+                self._query.update({
```
mparent61: delete me?

bulbs/special_coverage/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    Reference to all SpecialCovearge queries.
+
+    :param only_recent: references RECENT_SPONSORED_OFFSET from django settings.
+        Used to return sponsored content within a given configuration of hours.
+    :returns: Djes.LazySearch query matching all active speical coverages.
+    """
+    special_coverages = get_sponsored_special_coverages()
+    es_query = SearchParty(special_coverages).search()
+    if only_recent:
+        offset = getattr(settings, "RECENT_SPONSORED_OFFSET", 0)
```
mparent61: i'd add "_HOURS" to the variable name so units are explicit

tests/special_coverage/test_spec_cov_search.py
```diff
@@ -2,16 +2,75 @@
...
+        Content.search_objects.refresh()
+
+    def test_register_queryset_default_first(self):
+        queryset = Content.search_objects.search(feature_types=[self.feature_type1.slug])
+        reading_list = ReadingListIterator()
+        # First queryset should be default until explicitly state otherwise.
+        reading_list.register_queryset(queryset)
+        self.assertEqual(reading_list.default_queryset, queryset)
+        out = [obj for obj in reading_list]
+        for obj in queryset.full():
```
mparent61: the problem with this approach is that if queryset ever returns unexpected results (maybe ES mapping changes, ES request silently fails and no results returned, etc.), the testing assertions will never run. You could add a length check, but better to just be explicit and compare with an array of the exact expected results:

`self.assertEqual(out, [self.content[0], self.content[10], ....])`

bulbs/special_coverage/search.py
```diff
@@ -1,27 +1,110 @@
...
+        self.index = 0
+
+    def __iter__(self):
+        return self
+
+    def next(self):
+        for validator, queryset in self.querysets.items():
+            if validator(self.index):
+                self.index += 1
+                return queryset.next()
```
mparent61: should handle `StopIteration`, probably just ignoring and continuing to next queryset

bulbs/special_coverage/search.py
```diff
@@ -1,27 +1,110 @@
...
+
+                })
+                self._query["excluded_ids"] += query.get("excluded_ids", [])
+                self._query["included_ids"] += query.get("included_ids", [])
+                self._query["pinned_ids"] += query.get("pinned_ids", [])
+                self._query["groups"] += [query.get("groups", [])]
+        return self._query
+
+
+class ReadingListIterator(object):
```
mparent61: Good class!

benghaziboy:
@mparent61 updated and whatnot

mparent61:
Otherwise :+1:  
Looking forward to wiring this up to video recirc too

bulbs/special_coverage/search.py
mparent61: Maybe this needs a more generic name, as it could potentially be used for non-reading list stuff

benghaziboy:
ok @mparent61 a few final changes after testing this in my onion env

bulbs/special_coverage/search.py
```diff
@@ -1,27 +1,122 @@
...
+        self.default_queryset = None
+        self.querysets = OrderedDict()
+        self.index = 0
+        self.limit = kwargs.get("limit", getattr(settings, "READING_LIST_LIMIT", 100))
+
+    def __iter__(self):
+        return self
+
+    def next(self):
+        if self.index > self.limit:
```
mparent61: shouldn't this be `>=`?
If limit == 1, this won't be TRUE until we've returned 2 items (and index == 2)

PR #125, opened 2016-02-18
Better percolator logging.

mparent61:
Passing 'exc' to logger.exception seems to raise an inner exception in production.

@MichaelButkovic @benghaziboy


bulbs/__init__.py
```diff
@@ -1 +1 @@
-__version__ = "0.7.4"
+__version__ = "0.7.5"
```
mparent61: Cam I'm stealing your .5

benghaziboy:
yeah sure whatever

PR #127, opened 2016-02-25
Special Coverage view

MichaelButkovic:


bulbs/special_coverage/api.py
```diff
@@ -7,7 +7,6 @@
 
 
 class SpecialCoverageViewSet(viewsets.ModelViewSet):
-    queryset = SpecialCoverage.objects.all()
```
mparent61: Just curious, what are advantages of moving to inside `get_queryset()` ?

bulbs/special_coverage/views.py
```diff
@@ -0,0 +1,44 @@
...
+        return qs[0]
+
+    def get_context_data(self, *args, **kwargs):
+        context = super(SpecialCoverageView, self).get_context_data()
+        context["content_list"] = self.special_coverage.get_content()
+
+        return context
+
+
+special_coverage = SpecialCoverageView.as_view()
```
mparent61: should we wrap inside caching?

tests/special_coverage/test_spec_cov_views.py
```diff
@@ -0,0 +1,109 @@
...
+            description="Testing special coverage",
+            query={
+                "included_ids": [content.id]
+            },
+            videos=[],
+            campaign=self.campaign,
+            start_date=timezone.now() - timezone.timedelta(days=20),
+            end_date=timezone.now() - timezone.timedelta(days=10)
+        )
+        sc.save()
```
mparent61: Just for my own Django knowledge, do you need the explicit `save()` here?

spra85:
:+1: - So, the implementation on this is as simple as changing the `special/` url view in each of the properties and bumping `django-bulbs`?

tests/special_coverage/test_spec_cov_views.py
```diff
@@ -0,0 +1,109 @@
...
+            videos=[],
+            campaign=self.campaign,
+            start_date=timezone.now() - timezone.timedelta(days=10),
+            end_date=timezone.now() + timezone.timedelta(days=10)
+        )
+        sc.save()
+
+        response = self.client.get(reverse("special", kwargs={"slug": sc.slug}))
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.template_name[1], "special/landing.html")
```
mparent61: what about `template_name[0]`?

bulbs/special_coverage/api.py
MichaelButkovic: ah, i was just trying something out, so i can put this back. theoretically, using `get_queryset()` would be useful if we wanted to dynamically adjust the queryset.

bulbs/special_coverage/views.py
spra85: Good catch, we should unless that's something that would happen in each property?

mparent61: How about a default, that could be overridden in settings? This seems like something easy to forget otherwise

tests/special_coverage/test_spec_cov_views.py
MichaelButkovic: we actually don't need this either. i was running into some unrelated elasticsearch issues and forgot to take this out.

bulbs/special_coverage/views.py
```diff
@@ -0,0 +1,44 @@
...
+class SpecialCoverageView(BaseContentDetailView):
+    redirect_correct_path = False
+
+    def show_published_only(self):
+        return bool("full_preview" not in self.request.GET)
+
+    def get_template_names(self):
+        template_names = ["special_coverage/default.html"]
+
+        extended = getattr(settings, 'SPECIAL_COVERAGE_LANDING_TEMPLATE', None)
```
benghaziboy: ehhhhh. I don't like the pattern of having every template being configurable. We used it for the custom template logic since that varied a bit and is non-standard by nature, but this should be standard or the view should be overwritten

bulbs/special_coverage/views.py
```diff
@@ -0,0 +1,44 @@
...
+from django.shortcuts import get_object_or_404
+
+from bulbs.special_coverage.models import SpecialCoverage
+from bulbs.content.views import BaseContentDetailView
+
+
+class SpecialCoverageView(BaseContentDetailView):
+    redirect_correct_path = False
+
+    def show_published_only(self):
```
benghaziboy: move this up to BaseContentDetailView

bulbs/special_coverage/views.py
MichaelButkovic: hm, well since we're nixing proving a configurable template setting, i'd imagine most sites are going to be importing the view and extending from it, but i'm fine doing it any way.

benghaziboy: I think we're trying to avoid extending views

MichaelButkovic:
@spra85 yeah, then providing their own `special_coverage/landing.html` template

MichaelButkovic:
@mparent61

mparent61:
LGTM
@benghaziboy you happy with template config scheme?

mparent61:
@MichaelButkovic   update changelog too

PR #130, opened 2016-02-26
Contribution hotfix

benghaziboy:


benghaziboy:
@mparent61

bulbs/contributions/views.py
```diff
@@ -165,6 +165,7 @@ def list(self, request):
         format = self.request.QUERY_PARAMS.get('format', None)
         if format == 'csv':
             queryset = self.get_queryset()
+            queryset = ESPublishedFilterBackend().filter_queryset(request, queryset, None)
```
mparent61: Could there be any other future surprises from not calling super()?

mparent61:
Changelog!

mparent61:
:+1: With an appropriate and eloquent changelog

bulbs/contributions/views.py
benghaziboy: Possibly? I would like to rewrite this implementation so we aren't separating anything about the csv & api responses until the end, let that happen somewhere else.

TODO added.

PR #133, opened 2016-03-03
Vault client caching

mparent61:
@collin this is my thought on vault caching, still needs tests + such

### DIFF
--- a/bulbs/utils/vault.py
+++ b/bulbs/utils/vault.py
@@ -8,11 +8,17 @@
 import requests
 
 from django.conf import settings
+from django.core.cache import cache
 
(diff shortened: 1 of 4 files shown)


mparent61:
@collin @MichaelButkovic @benghaziboy 

Suggestion for Vault REST endpoint caching, what do you think?

Idea is to add 2 layers of caching:
- "short-term" caching to avoid overwhelming Vault service (~1 min)
- "backstop" caching (I made up the name) to fallback on if Vault endpoint is temporarily unreachable (~60 min)

Short-term seems like a no-brainer, the "backstop" concept could use some feedback.

benghaziboy:
niiiiiiice

PR #137, opened 2016-03-10
Cut default RSS caching from 10 to 5 min

mparent61:
For faster RSS updates, per social squad

Will also update on 0-7-x branch

@spra85 @benghaziboy @MichaelButkovic

PR #138, opened 2016-03-10
Cut default RSS caching from 10 to 5 min

mparent61:
For faster RSS updates, per social squad

Backporting https://github.com/theonion/django-bulbs/pull/137


benghaziboy:
same

mparent61:
whoops i screwed this one up

PR #139, opened 2016-03-11
Fix flaky evergreen test

mparent61:
If `make_content()` is not provided a content class type, it randomly chooses one, so every so often it was choosing the WRONG one, which caused inflated query results.

Recommend requiring explicit class type param for `make_content()`

@benghaziboy


benghaziboy:
Just Great

mparent61:
I don't need your patronizing

PR #140, opened 2016-03-11
Fixed flaky API test

mparent61:
Another issue with make_content() randomly choosing a content type.

Also some lint.

@benghaziboy


benghaziboy:
Just really terrific

PR #141, opened 2016-03-11
Fix occasional ES 503 test errors

mparent61:
Caused by accessing ES indices before shard allocation completed. Seems most apparent in test cases that don't call a "refresh" on search objects, like token auth tests. 

Reference: http://stackoverflow.com/a/30311714

Sample Error: https://travis-ci.org/theonion/django-bulbs/jobs/115362885

Was able to recreate + verify fix locally. Travis builds seem fixed too. 

Also threw in some lint fixes (see comments).


mparent61:
Also applies some lint + PY2/PY3 best practices

bulbs/utils/test/__init__.py
```diff
@@ -35,7 +34,7 @@ def make_content(*args, **kwargs):
         models = indexable_registry.families[Content]
         model_keys = []
         for key in models.keys():
-            if not key in ['content_content', 'poll_poll']:
+            if key not in ['content_content', 'poll_poll']:
```
mparent61: Lint

bulbs/utils/test/__init__.py
```diff
@@ -54,7 +53,7 @@ def default(self, value):
         iso = _iso_datetime(value)
         if iso:
             return iso
-        if not PY3 and isinstance(value, str):
+        if PY2 and isinstance(value, str):
```
mparent61: PY3+ best practice, better to check PY2, so we don't break on future PY4 (which will be more similar to 3 than 2)

mparent61:
@benghaziboy @MichaelButkovic @collin

MichaelButkovic:
:+1: very nice

PR #142, opened 2016-03-15
Series detail view

MichaelButkovic:
this adds a video app & creates a series detail view within it. similar to special coverage, we’ll just need to add the url and the overridden template site side. @mparent61 @benghaziboy


bulbs/videos/views.py
```diff
@@ -0,0 +1,44 @@
+from django.conf import settings
+from django.http import Http404
+from django.views.generic import DetailView
+from django.views.decorators.cache import cache_control
+
+import requests
+
+
+class SeriesDetailView(DetailView):
+    redirect_correct_path = False
```
mparent61: Why is this set?

mparent61:
:+1:

bulbs/videos/views.py
MichaelButkovic: ah, yeah, this can be removed now. it was a holdover from using `BaseContentDetailView`.

PR #145, opened 2016-03-17
Recirc

MichaelButkovic:
opening this up early so people can take a look @mparent61 @benghaziboy @spra85


bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def save(self, *args, **kwargs):
+        self.clean()
+
+        super(InlineRecircMixin, self).save(*args, **kwargs)
+
+        if self.query and self.query != {}:
+            self._save_percolator()
+
+    def _save_percolator(self):
```
spra85: Do we need a percolator for this scenario?  With percolator, we use when we need to ask, is X item part of Y query.  But I didn't think we needed that scenario.  It's always give me a list of X items based on Y query

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,115 @@
...
+            q = self.query
+        search = custom_search_model(Content, q, published=published, field_map={
+            "feature_type": "feature_type.slug",
+            "tag": "tags.slug",
+            "content-type": "_type"
+        })
+        return search
+
+    @property
+    def es_id(self):
```
benghaziboy: -percolator

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,115 @@
...
+            q['included_ids'] = self.query.get('included_ids', [])
+
+        es.index(
+            index=index,
+            doc_type=".percolator",
+            body=q,
+            id=self.es_id
+        )
+
+    def _delete_percolator(self):
```
benghaziboy: -percolator

benghaziboy:
Fine. I would test this with a new test object in `example/testcontent/models.py` that uses it so we can test basic migration/save/validation

MichaelButkovic:
yeah, i was just in the middle of building out the tests with a new object

MichaelButkovic:
@mparent61 @benghaziboy @spra85 going to find a sane way to reverse the order of the content objects in the response, but other than that, this should be good for review

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,86 @@
...
+                    self.query[key] = [v for v in value if v is not None]
+
+    def save(self, *args, **kwargs):
+        self.clean()
+        super(BaseQueryMixin, self).save()
+
+    def get_recirc_content(self, published=True):
+        """gets the first 3 content objects in the `included_ids`
+        """
+        if "query" in self.query:
```
benghaziboy: turn this into a method on the object

benghaziboy:
outside of the reusable method, lgtm

bulbs/recirc/mixins.py
mparent61: good idea

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def clean(self):
+        super(BaseQueryMixin, self).clean()
+        self.clean_query()
+
+    def clean_query(self):
+        """
+        Removes any `None` value from an elasticsearch query.
+        """
+        if self.query and self.query != {}:
```
mparent61: second "and" clause is redundant (first will fail if equals empty dict `{}`)

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,86 @@
...
+        })
+        return search
+
+    @property
+    def contents(self):
+        """performs .get_content() and caches it in a ._content property
+        """
+        if not hasattr(self, "_content"):
+            self._content = self.get_content()
+        return self._content
```
mparent61: This is OK, but we should look into using some kind of memoizing decorator. 

I used Pyramid's `reify` in a past life: https://github.com/Pylons/pyramid/blob/master/pyramid/decorator.py#L4

bulbs/recirc/views.py
```diff
@@ -0,0 +1,29 @@
...
+
+class RecircViewSet(views.APIView):
+
+    permission_classes = (AllowAny,)
+
+    def get(self, request, pk):
+        content = get_object_or_404(Content, id=pk)
+
+        if not content.is_published:
+            return Response(status=status.HTTP_400_BAD_REQUEST)
```
mparent61: Should this be a 404?

MichaelButkovic: Yeah, it could be, if you wanted. I guess 404 would be better for security purposes.

mparent61: Let's stay consistent with what we do for other unpublished stuff -- I think it's 404

mparent61:
:+1:

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,91 @@
...
+        self.clean()
+        super(BaseQueryMixin, self).save()
+
+    def get_recirc_content(self, published=True):
+        """gets the first 3 content objects in the `included_ids`
+        """
+
+        # NOTE: set included_ids to just be the first 3 ids,
+        # otherwise search will return last 3 items
+        if "query" in self.query:
```
MichaelButkovic: @mparent61 @benghaziboy updated this part to set explicitly set `included_ids` to the first 3 ids in the query, since `custom_model_search` was only returning the last 3 objects

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+
+        # NOTE: set included_ids to just be the first 3 ids,
+        # otherwise search will return last 3 items
+        if self.has_query:
+            q = self.query["query"]
+            q['query']['included_ids'] = q['query']['included_ids'][:3]
+        else:
+            q = self.query
+            q['included_ids'] = q['included_ids'][:3]
```
mparent61: Seems dangerous to modify the actual field data, for a read-only operation

bulbs/recirc/mixins.py
```diff
@@ -0,0 +1,87 @@
...
+        super(BaseQueryMixin, self).save()
+
+    def get_recirc_content(self, published=True):
+        """gets the first 3 content objects in the `included_ids`
+        """
+
+        # NOTE: set included_ids to just be the first 3 ids,
+        # otherwise search will return last 3 items
+        q = self.get_query()
+        q['included_ids'] = q['included_ids'][:3]
```
mparent61: (in case this comment was lost) -- Seems dangerous to modify the actual model field in an otherwise read-only operation (or is this commonly done with Django?)

MichaelButkovic: `self.query()` seems to return a copy of the model field. I just added a test asserting that it wasn't changed.

MichaelButkovic: Ah, nevermind! It actually modifies the field, yet since it isn't saved, the field isn't changed in the test. I'll change that to make a deep copy instead.

mparent61: :+1:

PR #147, opened 2016-03-24
move some code out of clickhole

collin:
@MichaelButkovic starting to move this out of clickhole into bulbs.

### DIFF
--- a/tests/special_coverage/test_spec_cov_views.py
+++ b/tests/special_coverage/test_spec_cov_views.py
@@ -45,7 +45,6 @@ class TestSpecialCoverageViews(BaseIndexableTestCase):
         self.assertEqual(response.context['special_coverage'], sc)
         self.assertEqual(response.context['content_list'].count(), sc.get_content().count())
         self.assertEqual(response.context['content_list'][0].id, content.id)
-        self.assertEqual(response.context['recirc'], None)
         self.assertEqual(response.context['targeting'], {
             'dfp_specialcoverage': 'test-coverage',
             'dfp_campaign_id': self.campaign.id,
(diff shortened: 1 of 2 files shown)


tests/special_coverage/test_spec_cov_views.py
```diff
@@ -45,6 +45,11 @@ def test_special_coverage_view(self):
         self.assertEqual(response.context['special_coverage'], sc)
         self.assertEqual(response.context['content_list'].count(), sc.get_content().count())
         self.assertEqual(response.context['content_list'][0].id, content.id)
+        self.assertEqual(response.context['recirc'], None)
```
collin: This assertion is wrong.

The value here is an instance of: `djes.search.LazySearch`.

How strickt of validation should I go for here?

collin:
See https://github.com/theonion/clickhole/pull/191

tests/special_coverage/test_spec_cov_views.py
collin: @MichaelButkovic I'm not sure exactly what I should be testing for in the result here.

Should I make some content and check for it?

I'm not quite sure how to set that test up.

MichaelButkovic: Hm, yeah, I'd say add some content to it and then just assert that `recirc` doesn't have any unpublished data and that it doesn't have itself.

collin:
@mparent61 or @benghaziboy can I get a 2nd here?

MichaelButkovic:
:+1:

mparent61:
:+1:

PR #149, opened 2016-03-29
Add tag based recirc fallback

MichaelButkovic:
@mparent61 @benghaziboy


bulbs/recirc/mixins.py
```diff
@@ -41,13 +41,25 @@ def save(self, *args, **kwargs):
     def get_recirc_content(self, published=True, count=3):
         """gets the first 3 content objects in the `included_ids`
         """
+        query = self.get_query()
+
+        # check if query has included_ids & if there are any ids in it
+        try:
+            if len(query['included_ids']) == 0:
+                raise KeyError
```
mparent61: Don't need a try/except block just to catch a failed "if" check. Instead you could just do something like:
```
if not query.get('included_ids'):
     .... Logic without included_ids ....
     return qs[:count]
.... 

Also, why the conditional behavior if no included_ids -- at least toss a comment in to explain.

Comment (MichaelButkovic) [inline bulbs/recirc/mixins.py]: ah, yeah. my thought for both cases is that it could have just been constructed and not have an `included_ids` array in the query, or someone could have removed all of the `included_ids` from it.

File: bulbs/recirc/mixins.py [hunk 2]
```diff
@@ -43,11 +43,9 @@ def get_recirc_content(self, published=True, count=3):
...
         query = self.get_query()
 
-        # check if query has included_ids & if there are any ids in it
-        try:
-            if len(query['included_ids']) == 0:
-                raise KeyError
-        except KeyError:
+        # check if query has included_ids & if there are any ids in it,
+        # in case the ids have been removed from the array
+        if not query.get('included_ids') or len(query.get('included_ids')) == 0:
```
mparent61: This works, but the len() check is redundant (if no included_ids, then "query.get('included_ids')" will be False.

MichaelButkovic: ah, perfect.

MichaelButkovic: this also exposed a bug where the master object would show up in the result sometimes

mparent61:
:+1:

PR #162, opened 2016-04-26
WIP - Recirc Widget and Styles app

jmelvnsn:


bulbs/styles/static/styles/stylesheets/mixins/flex-grid.scss
```diff
@@ -0,0 +1,38 @@
...
+// ************************* //
+///////////////////////////////
+
+@import 'flexbox';
+
+// taken from https://ajy.co/the-simplest-sass-flexbox-grid-ever
+
+// Unitless, interpreted as pixels
+
+$_fg_gutter: 24 !default; // gutter between columns. Set as desired.
```
collin: Every time this value is used it is multiplied by `1px`.

We should alter this to just accept a `#px` value.

I think we ought to clean up the argument names and accept actual values here. Then also encourage using the named arguments in practice:

```sass
@include flexgrid($columns: 4, $gutter: 24px, $padding: 10px);
```
vs
```sass
@include flexgrid(4, 24, 10);
```

PR #167, opened 2016-04-29
Fix Content API Trash using new DJES delete-index-on-save

mparent61:
Paired with https://github.com/theonion/djes/pull/29

Originally was treating ES 404 error as failure. There's no reason to error out if what you're trying to delete is already absent.

@benghaziboy @MichaelButkovic 

Will need to coordinate versioning w/ @kand


benghaziboy:
🔔

bulbs/api/views.py
```diff
@@ -175,30 +175,21 @@ def publish(self, request, **kwargs):
...
         :param kwargs: keyword arguments (optional)
         :return: `rest_framework.response.Response`
-        :raise: 404
         """
         content = self.get_object()
 
         content.indexed = False
         content.save()
 
-        index = content.__class__.search_objects.mapping.index
```
mparent61: delete has been moved into `Indexable.save()`

benghaziboy:
lol 👍

PR #171, opened 2016-05-03
Use campaign-display in special coverage landing

daytonn:
Added the `campaign-display` bulbs element to the special coverage landing page. This is pretty stinky since the dependencies for `bulbs-elements` are downstream. It was decided that at this time, we're just going to punt for now and revisit this when we have a better idea of how we want these dependencies to work.
 
@theonion/developers


spra85:
👍

collin:
👍

PR #191, opened 2016-05-19
Default feature type rates

daytonn:


daytonn:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;o7CUPRIK&#x2F;36-payroll-new-feature-type"> PayRoll: New Feature Type</a>

bulbs/content/models.py
```diff
@@ -117,6 +117,9 @@ def __unicode__(self):
         """
         return self.name
 
+    def is_new(self):
+        return not bool(self.pk)
```
kand: Not sure what the empty value is here, but I think you want to do something along the lines of:
```py
return self.pk is None
```

tests/content/test_content_tasks.py
```diff
@@ -0,0 +1,19 @@
...
+from bulbs.contributions.models import ContributorRole, FeatureTypeRate
+
+
+class FeatureTypeTasksTestCase(BaseIndexableTestCase):
+    def test_update_feature_type_rates(self):
+        role = ContributorRole.objects.create(name='Test Role', payment_type=1)
+        feature_type = FeatureType.objects.create(
+            name='Test Feature Type',
+            slug='test-feature-type',
+            instant_article=False)
```
kand: Nitpicky and not sure what the prevailing style is here, but I think these closing parens should go on the next line.

bulbs/content/models.py
daytonn: does `self.pk is None` return a boolean?

tests/content/test_content_tasks.py
daytonn: I'll adhere to whatever the prevailing style is here. I assumed that since python has no end of block syntax that indentation would be sufficient, so I used lisp style.

bulbs/content/models.py
```diff
@@ -117,6 +117,9 @@ def __unicode__(self):
         """
         return self.name
 
+    def is_new(self):
```
benghaziboy: just a preference, but I like to make these kinds of methods properties.

```
@property
def is_new(self):
    return self.pk is None
```

referenced as follows
```
bool(instance.is_new) "True"
```

bulbs/content/tasks.py
```diff
@@ -34,3 +33,21 @@ def index_feature_type_content(featuretype_pk):
...
+    from bulbs.contributions.models import ContributorRole, Rate, FeatureTypeRate
+
+    roles = ContributorRole.objects.all()
+
+    for role in roles:
+        existing_rates = FeatureTypeRate.objects.filter(
+            feature_type_id=featuretype_pk,
+            role_id=role.pk)
+
+        if existing_rates.count() == 0:
```
benghaziboy: doesn't mater but this could alternatively be written as 
```if existing_rates.exists()``` and queries might even be false-y if they are empty naturally.

File: bulbs/content/tasks.py [hunk 5]
```diff
@@ -34,3 +33,21 @@ def index_feature_type_content(featuretype_pk):
     featuretype = FeatureType.objects.get(pk=featuretype_pk)
     for content in featuretype.content_set.all():
         content.index()
+
+
+@shared_task(default_retry_delay=5)
+def update_feature_type_rates(featuretype_pk):
+    from bulbs.contributions.models import ContributorRole, Rate, FeatureTypeRate
+
+    roles = ContributorRole.objects.all()
```
benghaziboy: I think we want this query to only add it for roles that are "FeatureType" specific

so 

```
from bulbs.contribution.models import FEATURETYPE
...
...
roles = ContributorRole.objects.filter(payment_type=FEATURETYPE)
```

tests/content/test_content_tasks.py
benghaziboy: we use the flake8 linter (yes @mparent61?) generally adhering to pep8

tests/content/test_content_tasks.py
```diff
@@ -0,0 +1,19 @@
+from bulbs.content.models import FeatureType
+from bulbs.content.tasks import update_feature_type_rates
+from bulbs.utils.test import BaseIndexableTestCase
+from bulbs.contributions.models import ContributorRole, FeatureTypeRate
+
+
+class FeatureTypeTasksTestCase(BaseIndexableTestCase):
```
benghaziboy: Let's test to make sure an arbitrary save on an existing featuretype won't create a duplicate as well

tests/content/test_content_tasks.py
daytonn: I have the linter activated in my editor. It has no problems with the syntax, what do people prefer?

bulbs/content/tasks.py
daytonn: For some reason, that did not work for me.

tests/content/test_content_tasks.py
mparent61: Yes flake8. We need to get django-bulbs fully lint-conforming ASAP so we can enforce linting in Travis.

bulbs/content/models.py
```diff
@@ -126,10 +130,17 @@ def save(self, *args, **kwargs):
...
+
         feature_type = super(FeatureType, self).save(*args, **kwargs)
+
         if self.instant_article_is_dirty:
             index_feature_type_content.delay(self.pk)
+
         self._db_instant_article = self.instant_article
+
+        if self.is_new:
+            update_feature_type_rates(self.pk)
```
benghaziboy: overlooked this. 

celery tasks should be called as follows
```update_feature_type_rates.delay(self.pk)```

File: bulbs/content/tasks.py [hunk 8]
```diff
@@ -34,3 +33,21 @@ def index_feature_type_content(featuretype_pk):
     featuretype = FeatureType.objects.get(pk=featuretype_pk)
     for content in featuretype.content_set.all():
         content.index()
+
+
+@shared_task(default_retry_delay=5)
+def update_feature_type_rates(featuretype_pk):
+    from bulbs.contributions.models import ContributorRole, FeatureTypeRate, FEATURETYPE
```
spra85: Is this non-standard to put import statements in method declarations like this? // @benghaziboy @MichaelButkovic

docs/conf.py
```diff
@@ -129,44 +130,44 @@
 
 # If not '', a 'Last updated on:' timestamp is inserted at every page bottom,
 # using the given strftime format.
-#html_last_updated_fmt = '%b %d, %Y'
+# html_last_updated_fmt = '%b %d, %Y'
```
spra85: OCD. Ha.

spra85:
Other than the trivial comments above and implementing @benghaziboy's suggestion on that add'l test case, looks fine to me 👍

bulbs/content/tasks.py
MichaelButkovic: it's not standard, no, but sometimes you'll need it to get around circular dependencies. not sure if that's the case here though.

benghaziboy: it is. It's somewhat necessary for celery tasks however due to circular imports. Since tasks are commonly called by objects they intend to reference in separate threads. Low memory cost for that same reason however.

bulbs/content/tasks.py
benghaziboy: no worries there.

benghaziboy:
@spra85 he did in a newer test we went over. 👍 @MichaelButkovic

MichaelButkovic:
👍

docs/conf.py
daytonn: I just have an in-editor linter so I added it

PR #196, opened 2016-05-23
get all campaigns for tunic.

benghaziboy:
@mparent61 @MichaelButkovic 

Ran into an issue with the tunic client, which was only retrieving the 1st page of campaign data. When provided the `next` url the client would add the domain info back to the url & broke `get`


bulbs/utils/tunic.py
```diff
@@ -18,20 +18,37 @@ def get(self, path):
...
+        return urljoin(
+            url_base, settings.TUNIC_API_PATH.rstrip('/') + '/' + path.lstrip('/')
+        )
+
+    def get_campaigns(self, filter_active=False):
+        campaigns = {}
+        path = "/campaign/"
+        if filter_active:
+            path += "?active=True"
+        while True:
```
mparent61: could be ```while path:```

benghaziboy:
@mparent61 updated

MichaelButkovic:
👍

tests/utils/test_tunic.py
```diff
@@ -25,6 +25,35 @@ def test_get_active_campaigns(self):
                               2: {'id': 2,
                                   'name': 'Another Campaign'}})
 
+    def test_get_all_paginated_campaigns(self):
+        with requests_mock.mock() as mocker:
```
mparent61: very cool `requests_mock`

benghaziboy: I was actually writing something like 

```
class MockResponse(object):

    def __init__(self, content):
        self.content = content

    def json(self):
        return self.content```

on the onion until I saw `requests_mock` in bulbs :(

mparent61:
LGTM

PR #209, opened 2016-05-26
Instant Article API publishing

MichaelButkovic:


MichaelButkovic:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ylgOHMAf&#x2F;12-integration-with-fb-api"> Integration with FB API</a>

bulbs/content/models.py
```diff
@@ -597,7 +602,18 @@ def content_deleted(sender, instance=None, **kwargs):
         cls.search_objects.client.delete(index, doc_type, instance.id, ignore=[404])
 
 
+def delete_from_instant_article_api(sender, instance=None, **kwargs):
+    fb_access_token = vault.read()['value']
+    
+    if instance.instant_article_id:
+        requests.delete('https://graph.facebook.com/v2.6/{0}?access_token={1}'.format(
```
mparent61: URL should be a setting config, used everywhere

mparent61:
Looking good so far

MichaelButkovic:
@mparent61 @benghaziboy @spra85 ready for review

benghaziboy:
👍

bulbs/content/tasks.py
```diff
@@ -51,3 +52,93 @@ def update_feature_type_rates(featuretype_pk):
...
+                '{0}/{1}/instant_articles'.format(fb_api_url, fb_page_id),
+                data={
+                    'access_token': fb_access_token,
+                    'html_source': source,
+                    'published': 'true',
+                    'development_mode': 'false'
+                })
+
+            # Get status of article
+            status = requests.get('{0}/{1}?access_token={2}'.format(
```
spra85: I thought I read that status is not instantaneous and you almost have to poll - is that not the case?

bulbs/content/models.py
```diff
@@ -597,7 +602,31 @@ def content_deleted(sender, instance=None, **kwargs):
...
+        fb_api_url = getattr(settings, 'FACEBOOK_API_BASE_URL', '')
+        fb_access_token = vault.read(settings.FACEBOOK_TOKEN_VAULT_PATH)
+
+        delete = requests.delete('{0}/{1}?access_token={2}'.format(
+            fb_api_url,
+            instance.instant_article_id,
+            fb_access_token
+        ))
+
+        status = delete.json()["success"]
```
mparent61: use `.get('success')` to protect against missing value

mparent61: Also need to check `delete.ok` status before you access json(). On error code, `.json()` will raise.

bulbs/content/tasks.py
```diff
@@ -51,3 +52,93 @@ def update_feature_type_rates(featuretype_pk):
...
+
+
+@shared_task(default_retry_delay=5)
+def post_to_instant_articles_api(content_pk):
+    import requests
+    import logging
+
+    from django.conf import settings
+    from django.template.loader import render_to_string
+    from django.template.base import TemplateDoesNotExist
```
mparent61: can some of these be imported at the module level?

bulbs/content/tasks.py
```diff
@@ -51,3 +52,93 @@ def update_feature_type_rates(featuretype_pk):
...
+
+    from bulbs.utils import vault
+    from bulbs.instant_articles.renderer import InstantArticleRenderer
+    from bulbs.instant_articles.transform import transform
+    from .models import Content
+
+    if getattr(settings, 'FACEBOOK_API_ENV', '').lower() == 'production':
```
mparent61: This is a large code block that will be excluded from test + local environments. Does FB have a test server? Else I would minimize the scope of the "production only" portion -- perhaps only the POST + GET calls at the bottom (and in "non-production" environments you could just log the URL + payload for troubleshooting).

bulbs/content/tasks.py
```diff
@@ -51,3 +52,93 @@ def update_feature_type_rates(featuretype_pk):
...
+                source = render_to_string(
+                    "instant_article/_instant_article.html", context
+                )
+            except TemplateDoesNotExist:
+                source = render_to_string(
+                    "instant_article/base_instant_article.html", context
+                )
+
+            # Post article to instant article API
+            post = requests.post(
```
mparent61: need to check result (maybe just `post.ok`)

bulbs/content/tasks.py
```diff
@@ -51,3 +52,93 @@ def update_feature_type_rates(featuretype_pk):
...
+            if status.json()['status'] == 'SUCCESS':
+                # using a queryset.update() will stop save() from recursing
+                Content.objects.filter(pk=content_pk).update(instant_article_id=status.json()['id'])
+            else:
+                logger.error('Error in posting to Instant Article API: {}'.format(
+                    status.json()
+                ))
+        # if article is being unpublished, delete it from IA API
+        elif (feature_type and
+              feature_type.instant_article and
```
mparent61: Do you need these first 2 "feature type" conditions? Isn't it enough that the content has an "instant_article_id"? 

One bad situation would be if a particular feature type (say NIB) turned off IA, and then future NIB article deletes wouldn't remove the legacy IA content.

bulbs/instant_articles/templates/instant_article/base_instant_article.html
```diff
@@ -38,7 +38,7 @@ <h3 class="op-kicker">
 
         <figure>
           {% block cover %}
-            <img src="{% cropped_url content.detail_image.id %}" />
+            <img src="{% cropped_url content.detail_image.id fixed=True %}" />
```
mparent61: We should add a Tech Debt task to rename "fixed" and "BETTY_FIXED_URL" to something more expected like "force_public" and "BETTY_PUBLIC_IMAGE_URL`

tests/instant_articles/test_facebook_api.py
```diff
@@ -0,0 +1,149 @@
...
+        FACEBOOK_PAGE_ID='123456',
+        FACEBOOK_API_BASE_URL='https://graph.facebook.com/v2.6',
+        FACEBOOK_API_ENV='production',
+        FACEBOOK_TOKEN_VAULT_PATH='facebook/onion_token',
+        WWW_URL='www.theonion.com',
+        BETTY_FIXED_URL='http://i.onionstatic.com/onion'
+    )
+    @mock_vault({'facebook/onion_token': 'TOKEN'})
+    def test_publish(self):
+        with requests_mock.mock() as mocker:
```
mparent61: Nice mocking

tests/instant_articles/test_facebook_api.py
```diff
@@ -0,0 +1,149 @@
...
+            content.save()
+
+            Content.search_objects.refresh()
+            c = Content.objects.get(id=content.id)
+
+            # check that the ia_id is set & call count is correct
+            self.assertEqual(c.instant_article_id, 9876)
+            self.assertEqual(mocker.call_count, 2)
+
+    @override_settings(
```
mparent61: could these be set 1 time as class decorators (assuming they are the same for all 3 test cases)?

bulbs/content/tasks.py
```diff
@@ -51,3 +52,93 @@ def update_feature_type_rates(featuretype_pk):
...
+            # Get status of article
+            status = requests.get('{0}/{1}?access_token={2}'.format(
+                fb_api_url,
+                post.json()['id'],
+                fb_access_token
+            ))
+
+            # If upload if successful, set ia_id to id
+            if status.json()['status'] == 'SUCCESS':
+                # using a queryset.update() will stop save() from recursing
```
mparent61: Cool!

bulbs/content/tasks.py
MichaelButkovic: The API has a `development_mode` flag when posting, so maybe we should just set that with an environment variable. Part of the ticket was excluding this from local + testing, so I was just following that. @spra85 Any thoughts?

bulbs/content/tasks.py
MichaelButkovic: They were pretty instantaneous in my testing, but I can add polling here just in case.

tests/instant_articles/test_facebook_api.py
MichaelButkovic: class decorators weren't working for some reason (python 2 support?), but i was able to pair these down, so it's not as large.

MichaelButkovic:
@spra85 @mparent61 made updates based on feedback

bulbs/content/tasks.py
```diff
@@ -51,3 +63,117 @@ def update_feature_type_rates(featuretype_pk):
                 rate=0,
                 feature_type_id=featuretype_pk,
                 role_id=role.pk)
+
+
+@shared_task(default_retry_delay=5)
+def post_to_instant_articles_api(content_pk):
```
mparent61: This is a pretty huge function, i recommend breaking logic into 2 helper methods (add + delete)

bulbs/content/tasks.py
```diff
@@ -51,3 +63,117 @@ def update_feature_type_rates(featuretype_pk):
...
+                    IA ID: {1}\n
+                    Status Code: {2}'''.format(
+                        content.id,
+                        content.instant_article_id,
+                        post.status_code))
+                return
+
+            # Poll for status of article
+            response = ""
+            while response != "SUCCESS":
```
mparent61: Possible infinite loop if response never contains "success" or "error"  status -- any concern FB service might get stuck and keep returning a pending state? I'd include some kind of polling timeout to error out after say 5 minutes.

mparent61:
Remaining concern is that much of the FB API interaction is not covered in local/test environments, and only loosely covered by unit tests.

At the very least I would make sure that the settings access + template rendering is executed before the environment check, so that settings + template errors can be caught earlier in testing. 

What abou using VCR to record single IA article create + delete interactions, which could then be replayed via tests?

MichaelButkovic:
I'd be fine using VCR, but the requests mocks are already sending back the exact payloads from the API.

MichaelButkovic:
@mparent61 more updates

bulbs/content/tasks.py
```diff
@@ -51,3 +65,144 @@ def update_feature_type_rates(featuretype_pk):
...
+            FACEBOOK_PAGE_ID: {0}\n
+            FACEBOOK_API_BASE_URL: {1}\n
+            FB_ACCESS_TOKEN: {2}'''.format(
+                fb_page_id,
+                fb_api_url,
+                fb_access_token))
+        return
+
+    # render page source
+    context = {
```
mparent61: Yeah, this is good b/c it will execute this even in non-prod. However, currently rendering will happen even on deletes, which isn't necessary.

MichaelButkovic: ok, i just moved the ordering around so that it always renders when it should, but only posts when it's in prod

mparent61:
Yeah, the API responses are simple enough that VCR probably is overkill. VCR would potentially make maintenance easier in the future (auto-regenerate mocks), but we can punt on it. 

The refactoring you just did to always render even in non-prod seems like the low-hanging fruit here.

mparent61:
👍

bulbs/instant_articles/templates/instant_article/base_instant_article.html
MichaelButkovic: just added: https://trello.com/c/6dygMdoH/120-django-betty-cropper-rename-fixed-and-betty-fixed-url

mparent61: Sweet!

On Fri, Jun 3, 2016 at 1:13 PM, Michael Butkovic <notifications@github.com>
wrote:

> In
> bulbs/instant_articles/templates/instant_article/base_instant_article.html
> <https://github.com/theonion/django-bulbs/pull/209#discussion_r65742981>:
>
> > @@ -38,7 +38,7 @@ <h3 class="op-kicker">
> >
> >          <figure>
> >            {% block cover %}
> > -            
> > +            
>
> just added:
> https://trello.com/c/6dygMdoH/120-django-betty-cropper-rename-fixed-and-betty-fixed-url
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/theonion/django-bulbs/pull/209/files/90054b2101f00e3fa27534fce065f37fce2def2b#r65742981>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ADAkUblvhln6fUV6QTZunRK4i3t7R2XOks5qIGCogaJpZM4IoAZK>
> .
>



-- 
Mike Parent
Developer
Onion, Inc.

spra85:
👍

PR #210, opened 2016-06-02
Postgres testing

benghaziboy:
@mparent61 @MichaelButkovic 

Now that we are adding JSONB we must require postgres for environments. I think it'll be easiest for us to maintain this with our current docker config, so here we are.

The one issue is the python 2.7/3.5 testing cases, so we might have to test against 2 Dockerfiles? or install both into one container?


MichaelButkovic:
looks good

docker-compose.yml
```diff
@@ -0,0 +1,18 @@
+web:
+  build: .
+  environment:
+    DJANGO_SETTINGS_MODULE: example.settings
+  # FROM https://github.com/unbit/uwsgi-docker
```
mparent61: Don't need these 4 comments (confusing to leave them in)

example/settings.py
```diff
@@ -102,7 +106,7 @@
 
 ES_CONNECTIONS = {
     "default": {
-        "hosts": [os.environ.get('ELASTICSEARCH_HOST', 'localhost')],
+        "hosts": [os.environ.get('ELASTICSEARCH_HOST', 'elasticsearch')],
```
mparent61: This will break existing user setups if they are expecting to run "elasticsearch" on localhost, so you should note this in the README.

I've been hesitant to make this change because the wonderful Docker for Mac will make this unnecessary, as you can address docker containers via `localhost`.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+        libpq-dev \
+        postgresql-client-9.4 \
+        vim \
+    && rm -rf /var/lib/apt/lists/*
+
+# Deploy-specific python packages
+# - Django runs inside uWSGI
+# - docker-py required for Docker-assigned ephemeral port detection from within container
+RUN pip install "uwsgi>=2.0.11.1,<=2.1" \
+                "docker-py==1.4.0"
```
mparent61: Don't need either of these python libs (unless you are planning to run "example" app via UWSGI -- then you'd just need 'uwsgi')

mparent61:
I think newer docker allows you to pass arguments to the Dockerfile on build, so you could pass Python version 2.7/3.5.

mparent61:
Very nice to dockerize this!

PR #212, opened 2016-06-06
Use new attr param for host channel targeting

spra85:
@collin @MelizzaP @daytonn @kand ... This is the implemented change made possible through this bulbs-elements change (https://github.com/theonion/bulbs-elements/pull/85)


kand:
👍

collin:
👍

PR #214, opened 2016-06-06
Add additional facebook instant article settings

MichaelButkovic:
@spra85 @mparent61 @benghaziboy 
Added in these settings so we can test instant articles in development mode on testing


benghaziboy:
👍

bulbs/content/tasks.py
```diff
@@ -199,7 +201,9 @@ def post_to_instant_articles_api(content_pk):
                 source,
                 fb_page_id,
                 fb_api_url,
-                fb_token_path)
+                fb_token_path,
+                fb_dev_mode,
```
mparent61: Do you still need the `if environment == 'production':` checks?

MichaelButkovic: Ah, yeah that's true. I think so, because we don't want to send every article to development mode, we just want to use this for testing purposes. I think it also makes sense as an additional safety measure, but I'm fine with removing it as well.

mparent61: So what exactly do "fb_dev_mode" and "fb_publish" indicate? Maybe replace check with ```if fb_dev_mode or fb_publish``` or similar?

MichaelButkovic: `fb_dev_mode` indicates that if article should be sent to the Development Mode pane of the instant articles interface, and `fb_publish` indicates that if this article should be set to published when uploaded. We could check on `if not fb_dev_mode and fb_publish` but that seems sort of confusing to me.

mparent61: Agreed. But also confusing/error-prone to have to set `FACEBOOK_API_ENV = "production"` in order to test IA submissions even in a testing environment.
What about ```if environment == "production" or fb_dev_mode```?

MichaelButkovic: Yeah, that's true. Wouldn't that just always post then, assuming we're setting `fb_dev_mode` to True in base settings? Maybe we should nix the environment flag and just add a boolean `FACEBOOK_POST_TO_IA` flag, so we don't have to rely on these variables that we're posting to determine if we should send it or not?

mparent61: > Maybe we should nix the environment flag and just add a boolean FACEBOOK_POST_TO_IA flag

Agreed

MichaelButkovic:
@mparent61 updated

mparent61:
👍

PR #216, opened 2016-06-07
IA Vault key value pair

MichaelButkovic:
@mparent61 @benghaziboy vault apparently needs a key, along with the stored token value


bulbs/content/tasks.py
```diff
@@ -123,7 +123,7 @@ def post_article(content, body, fb_page_id, fb_api_url, fb_token_path, fb_dev_mo
 
 
 def delete_article(content, fb_api_url, fb_token_path):
-    fb_access_token = vault.read(fb_token_path)
+    fb_access_token = vault.read(fb_token_path)['authtoken']
```
benghaziboy: is there a risk of this raising an error and will that block anything?

MichaelButkovic: only if it's added to vault with the wrong key name

benghaziboy: I'd prefer that it does a `.get(authtoken, None)` so we can check and not fail.

mparent61: Best you can do is fail gracefully, by catching the error and raising exception to notify sentry.

MichaelButkovic: updated

bulbs/content/models.py
```diff
@@ -616,7 +616,7 @@ def delete_from_instant_article_api(sender, instance=None, **kwargs):
                     fb_token_path))
             return
 
-        fb_access_token = vault.read(fb_token_path)['authtoken']
+        fb_access_token = vault.read(fb_token_path).get('authtoken', None)
```
mparent61: If you do this, then you need to check for `None` and raise. 
Also you don't need `, None`, this is the default and makes for cleaner code
```
     fb_access_token = vault.read(fb_token_path).get('authtoken')
```

MichaelButkovic:
@mparent61 @benghaziboy updated

mparent61:
👍

PR #211, opened 2016-06-03
Initial videohub_client integration

mparent61:
Merging videohub_client.VideohubVideo support into Bulbs for standard Video interface

Each property's Video content models will inherit from new `VideoMixin`

Does not contain any breaking changes, as the updated Video filters aren't yet used by any properties.

PR #219, opened 2016-06-10
add more logging to IA requests

MichaelButkovic:
@mparent61 @benghaziboy


bulbs/content/models.py
```diff
@@ -632,10 +632,12 @@ def delete_from_instant_article_api(sender, instance=None, **kwargs):
...
                 Content ID: {0}\n
                 IA ID: {1}\n
-                Status Code: {2}'''.format(
+                Status Code: {2}
+                Request: {3}'''.format(
                     instance.id,
                     instance.instant_article_id,
-                    delete.status_code))
+                    delete.status_code,
+                    delete.__dict__))
```
benghaziboy: This might be really noisy. What are specifically looking to get out of this?

MichaelButkovic: just looking to get all of the request info right now as its 400ing. we can always turn it down later.

benghaziboy: welp if there is no other way then 👍

mparent61: You could add it as an additional `.debug` log, that can then be enabled per-property.
Up to you. LGTM
When you merge I will release this with v2.7.0 (https://github.com/theonion/django-bulbs/pull/211)

MichaelButkovic: I'll just merge this for now, since I imagine this will be a thing I'll remove or at least turn down once we figure out what error info is relevant.

PR #221, opened 2016-06-13
Streamline Travis builds

mparent61:
Quick wins to reduce Travis build matrix by 50%

- Drop Python 3.4 support (all sites use 2.7 or 3.5 now)
- Only run latest ES (1.7.5) with latest Python (3.5). This is a tradeoff for limited future-proofing.


mparent61:
@benghaziboy @MichaelButkovic @kand

MichaelButkovic:
👍

PR #222, opened 2016-06-13
Make instant_article_id a BigInteger

MichaelButkovic:


bulbs/videos/migrations/0002_auto_20160613_1116.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
```
MichaelButkovic: @mparent61 This file was generated when I generated my migrations. Do we want to add this?

MichaelButkovic:
@mparent61 @spra85 @benghaziboy related sentry error http://sentry.onion.com/onion/onion-test/issues/20461/

mparent61:
👍

bulbs/videos/migrations/0002_auto_20160613_1116.py
mparent61: Ah, I guess this new migration is needed after I reverted to the default ID behavior. Sure, let's add it

PR #223, opened 2016-06-13
Command to migrate Feature Type to IA

MichaelButkovic:

### DIFF
--- a/bulbs/content/management/commands/migrate_to_ia.py
+++ b/bulbs/content/management/commands/migrate_to_ia.py
@@ -1,6 +1,7 @@
 from django.core.management.base import BaseCommand
 
 from bulbs.content.models import Content, FeatureType
+from bulbs.content.tasks import post_to_instant_articles_api
 
 import timezone
 
@@ -27,5 +28,4 @@ class Command(BaseCommand):
                     published__lte=timezone.now())
 
                 for c in content:
-                    # save to trigger celery task to post to Instant Article API
-                    c.save()
+                    post_to_instant_articles_api.delay(c.id)
--- a/bulbs/content/management/commands/migrate_to_ia.py
+++ b/bulbs/content/management/commands/migrate_to_ia.py
@@ -1,129 +1,16 @@
-import time
-
-from django.core.management.base import BaseCommand, CommandError
-from django.conf import settings
-from django.template.loader import render_to_string
-from django.template.base import TemplateDoesNotExist
+from django.core.management.base import BaseCommand
 
 from bulbs.content.models import Content, FeatureType
-from bulbs.instant_articles.renderer import InstantArticleRenderer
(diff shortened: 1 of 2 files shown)


bulbs/content/management/commands/migrate_to_ia.py
```diff
@@ -0,0 +1,143 @@
...
+    fb_dev_mode = 'true' if getattr(settings, 'FACEBOOK_API_DEVELOPMENT_MODE', False) else 'false'
+    fb_publish = 'true' if getattr(settings, 'FACEBOOK_API_PUBLISH_ARTICLE', False) else 'false'
+    should_post = getattr(settings, 'FACEBOOK_POST_TO_IA', False)
+    fb_access_token = vault.read(getattr(settings, 'FACEBOOK_TOKEN_VAULT_PATH', None)).get('authtoken')
+    www_url = getattr(settings, 'WWW_URL', None)
+
+    def add_arguments(self, parser):
+        parser.add_argument('feature', nargs="+", type=str)
+
+    def generate_body(self, content):
```
mparent61: This should be some sort of helper method, shared by other IA logic

mparent61:
This copies lots of IA post logic from elsewhere, any way to re-use existing logic? Maybe trigger a `save()`?

MichaelButkovic:
@mparent61 ah, yeah, would work.

MichaelButkovic:
@mparent61 updated

mparent61:
LGTM

bulbs/content/management/commands/migrate_to_ia.py
```diff
@@ -0,0 +1,31 @@
...
+
+                # All published content belonging to feature type
+                content = Content.objects.filter(
+                    feature_type=ft,
+                    published__isnull=False,
+                    published__lte=timezone.now())
+
+                for c in content:
+                    # save to trigger celery task to post to Instant Article API
+                    c.save()
```
spra85: Seems like this is also unnecessarily triggering an ElasticSearch re-index?  Also, wouldn't it be possible for this script to overwrite somebody's change if it fetched the content and it was modified by the time this triggered a save?

mparent61: Good point. Maybe a "save to IA" helper method (also triggered by `save()`)?

MichaelButkovic: i was just thinking of just calling the celery task, but that doesn't solve the latter issue.

MichaelButkovic: actually, i think just calling the celery task will solve the issue, since it gets the content as the first thing in the task. if the content is updated after that, `save()` will be called, which should post the article to the API again.

spra85:
👍

mparent61:
👍

PR #225, opened 2016-06-14
Fix migrate to ia script

MichaelButkovic:
@spra85 @mparent61 @benghaziboy this fixes passing in the slug as an argument


mparent61:
👍

PR #227, opened 2016-06-15
Add optional AppConfigs as workaround for duplicate app names

mparent61:
Workaround for Django errors attempting to install 2 apps with same name (ex: `bulbs.videos` and `starwipe.apps.videos`). Projects can choose to explicitly include these AppConfigs with namespaced labels.

```
INSTALLED_APPS = [
   ...
   "bulbs.videos.apps.AppConfig",
   ...
]
```

And then refer to the app as `bulbs_videos`.

Note: I did not make these defaults (```default_app_config = 'bulbs.videos.apps.AppConfig'```) to avoid breaking existing integrations). Integrations must explicitly use `.app.AppConfig` as needed.


mparent61:
@benghaziboy @MichaelButkovic 

Also @MelizzaP can you test this with `bulbs.ads`?

benghaziboy:
lgtm. We'll want to decide if this is our long term solution however

MichaelButkovic:
LGTM

PR #234, opened 2016-06-23
Super Features

MichaelButkovic:
opening this up early @mparent61 @benghaziboy @spra85


mparent61:
Looking good

benghaziboy:
👍

bulbs/page/enum.py
```diff
@@ -0,0 +1,14 @@
+from django_enumfield import enum
+
+
+class Status(enum.Enum):
+    DRAFT = 0
+    PUBLISHED = 1
+
+
+class TemplateType(enum.Enum):
```
MichaelButkovic: @spra85 I just got these values from the mocks. Are there any other template types that we know of?

MichaelButkovic:

MichaelButkovic:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;iqjqr8ed&#x2F;33-1-super-feature-page-class-model"> (1) Super Feature Page Class&#x2F;Model</a>

MichaelButkovic:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;kXYUXcGF&#x2F;35-1-super-feature-page-serializer"> (1) Super Feature Page Serializer</a>

bulbs/super_features/enum.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class Status(enum.Enum):
+    DRAFT = 0
+    PUBLISHED = 1
+
+
+class TemplateType(enum.Enum):
+    GLOSSARY = 0
+    TWO_TYPES = 1
```
spra85: `TWO_TYPES` ?  Which one is that

MichaelButkovic:
@mparent61 @benghaziboy this is pretty much ready for review, if you guys want to start taking a look

mparent61:
I will crack this bad boy open in the AM :)

bulbs/super_features/metadata.py
```diff
@@ -0,0 +1,52 @@
...
+        return super(BaseSuperFeatureMetadata, self).determine_metadata(request, view)
+
+    def get_custom_metadata(self, serializer, view):
+        fields_metadata = dict()
+        if hasattr(serializer, "__call__"):
+            serializer_instance = serializer()
+        else:
+            serializer_instance = serializer
+        for field_name, field in serializer_instance.get_fields().items():
+            if isinstance(field, BaseSuperFeatureDataField):
```
mparent61: Could this entire function be bubbled up to `BaseSimpleMetaData`? Seems like the only difference between this one and Infographic implementation is which class it checks for on this line (`BaseSuperFeatureDataField`)?

bulbs/super_features/models.py
```diff
@@ -0,0 +1,79 @@
...
+    class Mapping(Content.Mapping):
+        data = field.Object()
+
+        class Meta:
+            # Necessary to allow for our data field to store appropriately in Elasticsearch.
+            # A potential alternative could be storing as a string., we should assess the value.
+            dynamic = False
+
+    @property
+    def is_parent(self):
```
mparent61: nice

bulbs/super_features/serializers.py
```diff
@@ -0,0 +1,50 @@
...
+
+    class Meta:
+        model = ContentRelation
+        fields = ('id', 'parent_id', 'child_id', 'ordering')
+
+
+class BaseSuperFeatureDataField(serializers.Field):
+
+    def to_internal_value(self, data):
+        serializer = get_data_serializer(
```
mparent61: I'd call this variable `serializer_class` to be consistent with `to_representation` below

bulbs/super_features/serializers.py
```diff
@@ -0,0 +1,50 @@
...
+            self.parent.initial_data.get("superfeature_type")
+        )
+        return serializer().to_internal_value(data)
+
+    def to_representation(self, obj):
+        serializer_class = get_data_serializer(
+            self.parent.initial_data.get("superfeature_type")
+        )
+        serializer = serializer_class(data=obj)
+        serializer.is_valid(raise_exception=True)
```
mparent61: Didn't know about `raise_exception`, nice

bulbs/super_features/utils.py
```diff
@@ -0,0 +1,64 @@
...
+    if serializer is None:
+        raise KeyError('The requested SuperFeature does not have a registered serializer')
+
+    return serializer
+
+
+def get_superfeature_model():
+    from bulbs.super_features.models import BaseSuperFeature
+
+    model_name = getattr(settings, "BULBS_SUPERFEATURE_MODEL", None)
```
mparent61: So the idea is that each property must set it's own model class here, similar to the new `settings.VIDEO_MODEL`, because it will inherit from property-specific base content?

bulbs/super_features/utils.py
```diff
@@ -0,0 +1,64 @@
...
+        GUIDE_TO_ENTRY: GuideToChildSerializer
+    }.get(name, None)
+    if serializer is None:
+        raise KeyError('The requested SuperFeature does not have a registered serializer')
+
+    return serializer
+
+
+def get_superfeature_model():
+    from bulbs.super_features.models import BaseSuperFeature
```
mparent61: Nit: Don't need to import this unless you're inside the `if` statement below

bulbs/super_features/viewsets.py
```diff
@@ -0,0 +1,35 @@
...
+    permission_classes = (IsAdminUser, CanEditContent,)
+
+    def get(self, request, pk):
+        child_ids = ContentRelation.objects.filter(
+            parent_id=pk
+        ).values_list(
+            'child_id', flat=True
+        ).order_by('ordering')
+        children = SUPERFEATURE_MODEL.objects.filter(id__in=child_ids)
+        result = SUPERFEATURE_PARTIAL_SERIALIZER(children, many=True)
```
mparent61: do you need to check if result.data `is_valid`?

bulbs/super_features/utils.py
MichaelButkovic: right, exactly.

bulbs/super_features/viewsets.py
MichaelButkovic: You only need to check `is_valid` if you're using the `data=` kwarg when instantiating the serializer. In this case, this will just return an empty list if there are no children.

tests/super_features/test_api.py
```diff
@@ -0,0 +1,150 @@
...
+            'parent_id': parent.id,
+            'child_id': child.id,
+            'ordering': 1
+        }
+        url = reverse('content-relation-list')
+        resp = self.api_client.post(url, data=json.dumps(data), content_type='application/json')
+
+        self.assertEqual(resp.status_code, 201)
+        self.assertTrue(child.is_child)
+        self.assertFalse(parent.is_child)
```
mparent61: I'd ad `is_parent` checks here too

bulbs/super_features/metadata.py
MichaelButkovic: i don't think so, because for the case of super features, each property will have its own`DataField`, along with it's own `metadata.py`

bulbs/super_features/models.py
```diff
@@ -0,0 +1,79 @@
...
+    def get_data_serializer(self):
+        return get_data_serializer(self.superfeature_type)
+
+    def validate_data_field(self):
+        Serializer = self.get_data_serializer()  # NOQA
+        Serializer(data=self.data).is_valid(raise_exception=True)
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        if self.is_child:
```
mparent61: We should discuss this

bulbs/super_features/metadata.py
mparent61: But what if there was a base method that took the `DataField` type as a parameter?

bulbs/super_features/models.py
```diff
@@ -0,0 +1,81 @@
...
+
+class BaseSuperFeature(Content, AbstractSuperFeature):
+    parent = models.ForeignKey('self', blank=True, null=True)
+    ordering = models.IntegerField(blank=True, null=True, default=None)
+
+    class Meta:
+        unique_together = ('parent', 'ordering')
+
+    class Mapping(Content.Mapping):
+        parent = field.Integer()
```
MichaelButkovic: @mparent61 @benghaziboy DJES would infinitely recurse if i didn't add this. not sure if this is terrible or not though.

mparent61: Any downsides besides this extra statement? 
If not, I'd just drop a comment in, if we do this enough we can add support to DJES

MichaelButkovic: I don't think there's any downside in this case, since the only time it will be not null, we don't want to index it anyway.

benghaziboy:
👍 don't merge today.

bulbs/content/models.py
```diff
@@ -367,7 +371,8 @@ def save(self, *args, **kwargs):
         """
         if not self.slug:
             self.slug = slugify(self.build_slug())[:self._meta.get_field("slug").max_length]
-        if self.indexed is False:
+
+        if self.is_indexed is False:
```
mparent61: `if not self.is_indexed:` is more pythonic

bulbs/super_features/utils.py
```diff
@@ -17,39 +18,51 @@ def get_data_serializer(name):
...
     serializer = getattr(settings, "BULBS_SUPERFEATURE_SERIALIZER", None)
-    if serializer is None:
+    if serializer is None or module is None:
+        from bulbs.super_features.serializers import BaseSuperFeatureSerializer
         return BaseSuperFeatureSerializer
-    return serializer
+    else:
+        # NOTE: import serializer module & then evaluate based on serializer name
+        mod = import_module(module)  # NOQA
+        return eval("{0}.{1}".format('mod', serializer))
```
mparent61: why is `eval` necessary here?

tests/super_features/test_api.py
```diff
@@ -54,6 +54,55 @@ def test_post_guide_to(self):
...
+        results = self.es.search(
+            BaseSuperFeature.search_objects.mapping.index,
+            BaseSuperFeature.search_objects.mapping.doc_type
+        )
+        self.assertEqual(results['hits']['total'], 1)
+
+        child = BaseSuperFeature.objects.get(id=child_id)
+        parent = BaseSuperFeature.objects.get(id=parent_id)
+
+        # CHECK IF PARENT AND ARE SETTING
```
mparent61: typo

bulbs/super_features/utils.py
MichaelButkovic: `eval()` is basically getting us the actual classes that we want. since we have the module and the string name of the class, we evaluate them together in dot form to get the class.

bulbs/super_features/utils.py
```diff
@@ -0,0 +1,77 @@
...
+def get_superfeature_serializer():
+    module = getattr(settings, "BULBS_SUPERFEATURE_SERIALIZER_MODULE", None)
+    serializer = getattr(settings, "BULBS_SUPERFEATURE_SERIALIZER", None)
+    if serializer is None or module is None:
+        from bulbs.super_features.serializers import BaseSuperFeatureSerializer
+        return BaseSuperFeatureSerializer
+    else:
+        # NOTE: import serializer module & then evaluate based on serializer name
+        mod = import_module(module)  # NOQA
+        return eval("{0}.{1}".format('mod', serializer))
```
mparent61: You don't need to use `eval` here. Use `getattr(mod, serializer)`

MichaelButkovic: ah, yeah, good call.

mparent61:
👍

PR #235, opened 2016-06-24
the greater good

benghaziboy:
making this an abstract baseclass, but left in the main model for the time being.

@mparent61 @MichaelButkovic


benghaziboy:
one thing that I noticed will be painful about the json fields, is that changing a serializer name will require some manual data migrations

tests/infographics/test_models.py
```diff
@@ -19,7 +19,7 @@ def test_create_success(self):
             infographic_type=InfographicType.LIST,
             data={
                 "is_numbered": True,
-                "items": [{
+                "objects": [{
```
mparent61: wouldn't these be called "entries"? It's also a bit less generic than "objects".

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Development
 
+## Version 3.1.0
+
+- Added `AbstractInfographic` model to refer to all infographic specific attributes.
+- Rename Infographic.data,items to Infographic.data.objects due to
```
mparent61: entries

mparent61:
LGTM

PR #237, opened 2016-06-28
infographic metadata expensive choice query

benghaziboy:
@mparent61 @MichaelButkovic 

this deals with queryset choices being an expensive query in our actual environments while not being useful., This is what I was referring to as being fixed in `rest_framework>=3.3`, but that route is also more complicated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Development
 
+## Version 3.1.1
+
+- Updated InfographicMetadata to exclude expensive `RelatedField` `ManyRelated` choice queries.
+
 ## Version 3.1.0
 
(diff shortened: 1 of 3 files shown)


mparent61:
Infographic logic is becoming quite complex. How much complexity goes away if we bite the bullet and upgrade DRF, and what is the effort involved?

benghaziboy:
Upgrading == breakages across multiple endpoints and idk how much requires refactor vs. minor update. 

This is basically a mirror of the current rest_framework logic for that one particular method.

benghaziboy:
but idk how we really avoid complexity here since we require a fair amount of custom structuring for our super features.

mparent61:
Alright 👍 
I would document what code can go away should we ever upgrade DRF

MichaelButkovic:
👍

PR #243, opened 2016-07-07
use videohub_ref id

collin:
@jmelvnsn @mparent61 Like this?


bulbs/homepage_player/templates/homepage_player/bulbs_homepage_player.html
```diff
@@ -59,7 +59,7 @@
           <bulbs-carousel-item
             {# TODO implement /v/2323 redirect route in urls.py #}
             {# see: https://trello.com/c/3bT7Xuc9/60 #}
-            share-url="http{% if request.is_secure %}s{% endif %}://{{ request.META.HTTP_HOST }}/v/{{ video.id }}"
+            share-url="{% build_video_share_uri videos.videohub_ref.id %}"
```
mparent61: "videos" shouldn't be plural

PR #258, opened 2016-07-25
Add Facebook image and description

MichaelButkovic:
@mparent61 @benghaziboy

### DIFF
--- a/bulbs/super_features/serializers.py
+++ b/bulbs/super_features/serializers.py
@@ -41,6 +41,10 @@ class BaseSuperFeatureSerializer(ContentSerializer):
     class Meta:
         model = BaseSuperFeature
 
+    def to_representation(self, obj):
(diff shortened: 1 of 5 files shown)


bulbs/super_features/serializers.py
```diff
@@ -41,6 +41,10 @@ class BaseSuperFeatureSerializer(ContentSerializer):
     class Meta:
         model = BaseSuperFeature
 
+    def to_representation(self, obj):
```
MichaelButkovic: just adding this for documentation purposes, since this serializer isn't being used

benghaziboy:
Just asking questions;

Do we maybe want a method for a fallback description? like if no facebook description -> regular description.

MichaelButkovic:
hm, not sure on that one. @spra85 any thoughts?

mparent61:
LGTM

PR #268, opened 2016-08-01
Gracefully handle only 1 video

daytonn:


spra85:
Seems fine, but leaning on @jmelvnsn / @collin to take a look since one of them implemented originally

collin:
We should take a look on test. But this seems reasonable to me.

collin:
👍

PR #272, opened 2016-08-03
Login redirect "next url" param preserves original query string

mparent61:
Currently query params are stripped when redirected to login page.

So if you go here:

```
/special/test?full_preview=true
```

After login you will be redirected back to 

```
/special/test
```

Losing the "full_preview" param.


mparent61:
@MichaelButkovic @benghaziboy   ready for review

MichaelButkovic:
LGTM

PR #274, opened 2016-08-04
Add super_feature list to SpecialCoverage model

MichaelButkovic:
@benghaziboy @mparent61 this creates a separate super_feature json list for SpecialCoverage


bulbs/special_coverage/migrations/0009_auto_20160208_1540.py
```diff
@@ -6,19 +6,11 @@
 from bulbs.special_coverage.models import SpecialCoverage
 
 
-def save_func(apps, schema_editor):
```
mparent61: I'd add a note for future maintainers about why this migration is now empty

bulbs/special_coverage/models.py
```diff
@@ -75,6 +77,13 @@ def clean_videos(self):
         if self.videos:
             self.videos = [int(v) for v in self.videos if v is not None and is_valid_digit(v)]
 
+    def clean_super_features(self):
+        """
+        Removes any null & non-integer values from the super feature list
+        """
+        if self.super_features:
+            self.super_features = [int(sf) for sf in self.super_features if sf is not None and is_valid_digit(sf)]  # NOQA
```
mparent61: why the #NOQA? 
If for too-long of a line, you can move "if" to second line

PR #275, opened 2016-08-05
Contribution report streaming

mparent61:
- CSV downloads now streamed, preventing timeouts. Previously 1-month queries would hit 60-second server (and possibly browser) timeouts.
- Added CSV filenames, and headers are now in fixed ordering (specified by custom Renderers)
- Also did some view cleanup.

## Test Links (via AVC)

http://fix-payroll-downloads.test.avclub.com/cms/app/reporting/


mparent61:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;on8QYNQW&#x2F;563-webtech-payroll-won-t-download"> WEBTECH - payroll won&#39;t download</a>

mparent61:
@benghaziboy  ready for review

benghaziboy:
outside of potentially duplicating the iterator logic, that I'm not sure of...so we shouldn't get hung up about. 👍

mparent61:
Maybe, but I don't know about it, and I can't find it in DJES. Eventually this should just be moved into DJES, but I'll pile that on as Tech Debt.

PR #281, opened 2016-08-15
Special Coverage optional end date

MichaelButkovic:
@mparent61 @benghaziboy


MichaelButkovic:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;k6SOPr2D&#x2F;542-make-special-coverage-end-date-optional"> Make special coverage end date optional</a>

mparent61:
There are potentially many filters that will require refactoring to support NULL end dates. Here's just a few from a quick search:

API Filtering: https://github.com/theonion/django-bulbs/blob/501c6a7486f1d0f92e3d8991a05ea574026c6f17/bulbs/api/views.py#L512

Percolator Filtering: https://github.com/theonion/django-bulbs/blob/master/bulbs/content/models.py#L508

bulbs/special_coverage/models.py
```diff
@@ -161,7 +159,9 @@ def get_content(self, published=True):
     @property
     def is_active(self):
         now = today_as_utc_datetime()
-        if self.start_date and self.end_date:
+        if self.start_date and not self.end_date:
```
mparent61: Only need to check `start_date` once:
```
if self.start_date:
    if self.end_date:
        return self.start_date <= now < self.end_date
    else: 
        return self.start_date <= now

File: bulbs/api/views.py [hunk 2]
```diff
@@ -509,7 +509,11 @@ def list(self, request):
                 active = get_query_params(self.request).get('active', '').lower()
                 now = timezone.now()
                 if active == 'true':
-                    qs = qs.filter(start_date__lte=now, end_date__gte=now)
+                    qs = qs.filter(
+                        start_date__lte=now, end_date__gte=now
+                    ) | qs.filter(
```
MichaelButkovic: @mparent61 decided to go with the union of the two querysets when updating all of these filters

bulbs/api/views.py
```diff
@@ -509,7 +509,11 @@ def list(self, request):
...
+                        start_date__lte=now, end_date__isnull=True
+                    )
                 elif active == 'false':
                     qs = qs.exclude(start_date__lte=now, end_date__gte=now)
```
mparent61: Sounds good. Don't forget "excludes" too.

bulbs/special_coverage/models.py
```diff
@@ -119,7 +120,9 @@ def _save_percolator(self):
         # support missing fields, so always need to include
 
         q["start_date"] = self.start_date
-        q["end_date"] = self.end_date
+        # NOTE: set end_date to datetime.max if special coverage has no end date
+        # (i.e. is a neverending special coverage)
+        q["end_date"] = self.end_date if self.end_date else datetime.max.replace(tzinfo=pytz.UTC)
```
mparent61: Nice

MichaelButkovic:
@mparent61 @benghaziboy

bulbs/special_coverage/filters.py
```diff
@@ -18,9 +18,13 @@ def filter_queryset(self, request, queryset, view):
             if value == "true":
                 queryset = queryset.filter(
                     start_date__lte=today_filter, end_date__gte=today_filter
+                ) | queryset.filter(
+                    start_date__lte=today_filter, end_date__isnull=True
```
benghaziboy: we _could_ do something like a custom manager query so we get 

`SpecialCoverage.objects.active()` rather than re-writing this a bunch of places, but that's mostly personal preference

benghaziboy:
I give it 💋

bulbs/special_coverage/filters.py
```diff
@@ -18,9 +18,13 @@ def filter_queryset(self, request, queryset, view):
...
                 )
             elif value == "false":
                 queryset == queryset.exclude(
                     start_date__lte=today_filter, end_date__gte=today_filter
+                ) | queryset.filter(
```
mparent61: Should be `.exclude()`

mparent61:
LGTM

bulbs/special_coverage/filters.py
mparent61: I like this idea

PR #282, opened 2016-08-15
needed to add data-videohub-base

shawncook:
hey @spra85 @collin @MelizzaP does this look correct to you? Still trying to figure out how to test it, but I'm pretty sure this'll fix [this bug](https://trello.com/c/z5aIZHbH/604-webtech-onion-hp-video-player-broken).

```
  if (typeof this.$videohubBaseUrl === 'undefined') {
    throw Error('CurrentEpisode requires bulbs-video to have a videohub base url.');
  }
```

☝️ i think this is what is causing the problem


collin:
I think the the script causing this bug shouldn't even be running on the home page.

And if it is going to run on the homepage, it shouldn't be selecting `$('bulbs-video')` so generally.

shawncook:
You're right. I made a new PR on onion here: https://github.com/theonion/onion/pull/535

PR #283, opened 2016-08-15
Data serializer register

benghaziboy:
@mparent61 @MichaelButkovic 

not ready yet.

Just looking for feedback.

Left it in bulbs, cause daddy wanted it there & it's currently light enough.


bulbs/serializer_registry/registry.py
```diff
@@ -0,0 +1,25 @@
...
+
+        from bulbs.serializer_registry import registry
+        print(registry.serializers)
+        [<class 'MySerializer'>]
+        """
+        def wrapped(cls):
+            name = kwargs.get('name', 'serializers')
+            registered = getattr(self.__class__, name, dict())
+            key = kwargs.get('key', cls.__name__)
+            registered[key] = cls
```
mparent61: Should validate that key doesn't already exist

tests/serializer_registry/test_registry.py
```diff
@@ -0,0 +1,37 @@
+from django.test import TestCase
+
+from bulbs.infographics import data_serializers
+from bulbs.infographics.enum import InfographicType
+from bulbs.serializer_registry import registry
+
+
+
+
+class RegistryTestCase(TestCase):
```
mparent61: I'd add a test case that creates a new serializer (inside the test def) w/ decorator, independent of Infographics

mparent61:
Looking good

bulbs/infographics/data_serializers.py
```diff
@@ -17,28 +20,34 @@ class ComparisonKeySerializer(serializers.Serializer):
     initial = serializers.CharField()
 
 
+# Top-level infographic serializers.
+@registry.serializer(name='infographics', key=InfographicType.COMPARISON)
```
mparent61: Maybe instead of generic `name`, should be `group`/`category` or similar?

benghaziboy: `path`?

mparent61: path feels too file-system-y. 
app? 
module?

benghaziboy: those feel like misnomers since they are both specific to django. I'll go with category

mparent61:
Replaced by https://github.com/theonion/django-bulbs/pull/285

PR #285, opened 2016-08-16
changed my damn mind

camsom:
ok. This is my final thought.

Both the constant/name/serializer need to be configured and there seems to be a fair amount of logic regarding how they are configured. This does it one place and replaces a lot of the utils logic.


bulbs/super_features/registry.py
```diff
@@ -0,0 +1,12 @@
+from bulbs.serializer_registry.registry import DataSerializerRegistry
+
+from .data_serializers import GuideToParentSerializer, GuideToChildSerializer
+
+
+GUIDE_TO_HOMEPAGE = 'GUIDE_TO_HOMEPAGE'
+GUIDE_TO_ENTRY = 'GUIDE_TO_ENTRY'
```
mparent61: Aren't these also defined in `models.py`?

benghaziboy: yeah. I will be moving them if we like this method.

mparent61:
I like this. Reminds me of the DRF router approach.

Why the need for separate `registry.py` module, could you just do the registration inside `data_serializers.py`?

benghaziboy:
Ostensibly? Might be some importing problems there///separate aims.

camsom:
See ya in hell

PR #288, opened 2016-08-18
change op-social to op-interactive in IA

MichaelButkovic:
Facebook changed `op-social` to `op-interactive` because they’re jerks @mparent61 @benghaziboy See the notice here https://developers.facebook.com/docs/instant-articles/reference/social


benghaziboy:
👍

mparent61:
HAHA lame

PR #291, opened 2016-08-22
add status filter to superfeature viewset

MichaelButkovic:
@mparent61 @benghaziboy


bulbs/super_features/filters.py
```diff
@@ -0,0 +1,22 @@
+import django_filters
+from rest_framework import filters
+
+from bulbs.super_features.utils import get_superfeature_model
+
+SUPERFEATURE_MODEL = get_superfeature_model()
+
+
+def filter_status(queryset, value):
```
MichaelButkovic: note: this isn't my favorite thing ever, but status is currently a property (that is different site side vs. bulbs side), so this seemed like an ok way to do this, unless we want to move status to being a model field for `Content`

mparent61:
LGTM (w/ chaining changes we discussed)

tests/super_features/test_viewsets.py
```diff
@@ -86,3 +88,11 @@ def test_super_feature_listing_search(self):
 
         self.assertEqual(len(response.data['results']), 1)
         self.assertEqual(response.data['results'][0]['title'], self.parent.title)
+
+    def test_super_feature_listing_filtering(self):
+        import pdb; pdb.set_trace()
```
benghaziboy: ----

mparent61: Agreed

MichaelButkovic: image

PR #294, opened 2016-08-23
Video redirect class

collin:
@mparent61 This passes the existing test cases.

My thinking is to extend this class in onion and override:

```py
def choose_video_for_redirect(self, videos):
```

To filter on the feature_type. ( in python, not as a db query )

Thoughts?


mparent61:
Yeah this is a good workaround. I would add plenty of comments that this will be removed once unique constraint is in place.

bulbs/videos/views.py
```diff
@@ -1,17 +1,37 @@
...
+        videos = self.get_videos_for_redirect(videohub_ref_id)
+        video = self.choose_video_for_redirect(videos)
+
+        if video is None:
+            raise Http404("No video found with the provided videohub_ref id.")
+
+        return HttpResponseRedirect(video.get_absolute_url())
+
+    def choose_video_for_redirect(self, videos):
+        try:
```
mparent61:
```
if videos:
    return videos[0]
```

bulbs/videos/views.py
```diff
@@ -1,17 +1,37 @@
...
+    def get(self, request, videohub_ref_id):
+        videos = self.get_videos_for_redirect(videohub_ref_id)
+        video = self.choose_video_for_redirect(videos)
+
+        if video is None:
+            raise Http404("No video found with the provided videohub_ref id.")
+
+        return HttpResponseRedirect(video.get_absolute_url())
+
+    def choose_video_for_redirect(self, videos):
```
mparent61: maybe rename to "temporary_workaround_to_choose_from_duplicate_videos()"

collin:
@benghaziboy @MichaelButkovic Lookin' for that second 👍

MichaelButkovic:
LGTM

PR #299, opened 2016-08-25
LiveBlog updates

mparent61:
- Add fields: `LiveBlog.pinned_content`, `LiveBlogResponse.internal_name`
- Fix LiveBlog model indexing (via `excludes`)
- Extra tests


example/testcontent/models.py
```diff
@@ -109,3 +109,8 @@ def get_absolute_url(self):
 
 class TestLiveBlog(Content, AbstractLiveBlog):
     """Example LiveBlog implementation"""
+
+    class Mapping(Content.Mapping):
```
mparent61: I suspect this mapping needs to be added to each concrete implementation (AVC, ON, etc) b/c DJES doesn't support multiple base class Mappings?

mparent61:
@benghaziboy @MichaelButkovic @kand   ready for review

example/testcontent/models.py
MichaelButkovic: yeah, this will overwrite each base contents mapping (`class Mapping(BaseContent.Mapping):`)

MichaelButkovic:
LGTM

PR #301, opened 2016-08-25
LiveBlog: Various updates

mparent61:
Building up list of minor LiveBlog improvements
- DELETE endpoint doc + tests
- `LiveBlogResponse` default ordering
- FIrebase entry update integration


bulbs/liveblog/tasks.py
```diff
@@ -0,0 +1,15 @@
...
+from django.utils import timezone
+
+
+@shared_task(default_retry_delay=5)
+def firebase_update_timestamp(liveblog_id):
+    endpoint = getattr(settings, 'LIVEBLOG_FIREBASE_NOTIFY_ENDPOINT', None)
+    if endpoint:
+        url = endpoint.format(liveblog_id=liveblog_id)
+        requests.put(url, json={
+            'updatedAt': timezone.now().timestamp()  # TODO: What time format?
```
mparent61: @kand @collin  -- What timestamp format to use? Currently this uses (float) seconds since epoch.

bulbs/liveblog/tasks.py
```diff
@@ -0,0 +1,16 @@
...
+from django.utils import timezone
+
+
+@shared_task(default_retry_delay=5)
+def firebase_update_timestamp(liveblog_id):
+    endpoint = getattr(settings, 'LIVEBLOG_FIREBASE_NOTIFY_ENDPOINT', None)
+    if endpoint:
+        url = endpoint.format(liveblog_id=liveblog_id)
+        resp = requests.put(url, json={
+            'updatedAt': timezone.now().timestamp()  # TODO: What time format?
```
collin: If this is the same date format as the GET entries api:

`/api/v1/liveblog/public/entry/`

`if_modified_since=2015-01-01T01:01:00.000Z`

Then the javascript can drop the the `updatedAt` value into the url.

(Assuming the `if_modified_since` filter is inclusive.)

mparent61:
@collin Updated Firebase `updatedAt` to use ISO date string (same as the "if_modified_since" param)

mparent61:
@benghaziboy @MichaelButkovic @collin 

Ready for review

MichaelButkovic:
LGTM

mparent61:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hbTk473L&#x2F;47-add-a-hook-for-firebase-in-the-backend"> Add a hook for firebase in the backend</a>

collin:
https://github.com/theonion/django-bulbs/blob/master/bulbs/liveblog/filters.py#L21

If we're going to be passing this value straight through, should that be `gte`? so we'll match the most recent update.

```
published__gte=when
```

mparent61:
@collin and I discussed a new solution (going to remove this JSON endpoint and add an HTML partial endpoint instead). I'll make a new PR

PR #302, opened 2016-08-26
Allow super feature child dates to be set for any parent publish date

kand:
The `set-children-dates` endpoint was using `parent.is_published` to check if it was a valid operation to set children publish dates. However, this flag is only `True` if the parent publish date is now or in the past. Changed the check to only check if publish date is set or not.


kand:
@MichaelButkovic @benghaziboy

kand:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;jRL6A0jj&#x2F;115-set-children-dates-endpoint-unable-to-set-dates-when-parent-publish-date-is-in-the-future"> Set children dates endpoint unable to set dates when parent publish date is in the future</a>

MichaelButkovic:
👍

benghaziboy:
👍

PR #305, opened 2016-08-29
Remove dependence on lodash

daytonn:


bulbs/special_coverage/static/special_coverage/js/special-coverage-loader.js
```diff
@@ -25,19 +23,21 @@ module.exports = (function () { // eslint-disable-line no-unused-vars
     this.defaultText = element.innerHTML;
     this.loadingText = 'Loading...';
 
-    defaults(options, {
+    options = Object.assign({
```
kand: I don't believe `Object.assign` is available in IE 11.

MelizzaP:
👍

bulbs/special_coverage/static/special_coverage/js/special-coverage-loader.js
```diff
@@ -1,9 +1,7 @@
 module.exports = (function () { // eslint-disable-line no-unused-vars
-  var _ = require('lodash');
-  var assign = _.assign;
-  var defaults = _.defaults;
-  var isUndefined = _.isUndefined;
-  var trim = _.trim;
+  function isUndefined (suspect) {
+    return suspect === undefined; // eslint-disable-line no-undefined
```
kand: I think it's better form to do `typeof(suspect) === 'undefined'`, avoids the eslint error probably as well.

daytonn: I actually just copied the implementation from lodash, I don't see any benefit in using typeof here.

daytonn: https://github.com/lodash/lodash/blob/4.15.0/lodash.js#L11880-L11882

kand: We'd have a real broken site if it were the case, but it avoids the edge case where `undefined` is reassigned. No reason to write more bulletproof code here when it's only a few more characters.

Lodash actually protects for this edge case in a different way here: https://github.com/lodash/lodash/blob/4.15.0/lodash.js#L11-L12

PR #310, opened 2016-09-01
Contributions: Fix failing tests + freezgun cleanup

mparent61:
This only broke on August 31, when adding 1 month to 8/31/2016 resulted in 9/31/2016, which is invalid (Sept only has 30 days).

@benghaziboy @MichaelButkovic


benghaziboy:
my bad
👍

mparent61:
This was a fun one to track down (seriously) :)

Somehow this one failure caused cascading errors on unrelated tests, I'm going to figure out why.

tests/contributions/test_contributions_api.py
```diff
@@ -951,11 +952,10 @@ def test_rolefield_rate_interpretation(self):
         contribution.delete()
 
 
+@freeze_time('2015-09-25')
 class ReportingApiTestCase(BaseAPITestCase):
     def setUp(self):
         super(ReportingApiTestCase, self).setUp()
-        self.freezer = freeze_time("2015-09-25")
```
mparent61: @benghaziboy 

So when these tests failed in setUp(), they caused lots of other unrelated tests to fail b/c they never ran tearDown and deactivated freezegun (`self.freezer.stop`). 

Decorator usage is better because it ensures cleanup.  However, this has a side-effect of causing `self.now` to equal the frozen time too ('2015-09-25') which I believe was the expectation?  (The class-level decorator means that the base `setUp` methods are included in the frozen time scope, including [assignment of `self.now`](https://github.com/theonion/django-bulbs/blob/master/bulbs/utils/test/__init__.py#L99)).

PR #314, opened 2016-09-02
[wip] Liveblog templates

collin:


bulbs/liveblog/templates/liveblog/bulbs_liveblog.html
```diff
@@ -0,0 +1,49 @@
...
+        </div>
+      </a>
+    </div>
+    {% endwith %}
+  {% endif %}
+</div>
+
+{% with entries=content.entries.all %}
+  {% liveblog_entries_partial %}
+{% endwith %}
```
collin: @mparent61 the partial endpoint should only have to render this templatetag with an array of entries in context.

PR #319, opened 2016-09-06
RSS excludes SuperFeatures

mparent61:
@MichaelButkovic @benghaziboy 

They won't render correctly anyway.


MichaelButkovic:
LGTM

PR #323, opened 2016-09-08
[wip] consolidated liveblog branches

collin:
consolidating liveblog branches

(getting difficult to work with split branches)


mparent61:
Sounds good 👍

PR #327, opened 2016-09-08
Styles add gradient mixin

daytonn:


shawncook:
🙌 Needed gradient and now we have it! Next up: box shadow, opacity, easing, transform/transition, border radius.

spra85:
👍

PR #328, opened 2016-09-08
We should probably run js tests

daytonn:


mparent61:
:) 👍

mparent61:
do you need to run `npm install` too, or is that automatically run by `npm test`?

daytonn:
whoops, yep

PR #330, opened 2016-09-09
Remove dependency on lodash again

daytonn:


MichaelButkovic:
👍  looks like javascript

PR #331, opened 2016-09-12
exclude super features from content list

MichaelButkovic:
@mparent61 @benghaziboy @kand we had talked about doing this instead of using the exclude query param, so i did it

### DIFF
--- a/bulbs/api/views.py
+++ b/bulbs/api/views.py
@@ -135,7 +135,13 @@ class ContentViewSet(UncachedResponse, viewsets.ModelViewSet):
             authors = get_query_params(self.request).getlist("authors")
             queryset = queryset.filter(Authors(authors))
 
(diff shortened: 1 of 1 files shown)


benghaziboy:
👍

kand:
So the `exclude` query param is now unused?

MichaelButkovic:
once this is merged, yeah. we can still include it if we want, but since it's only being used for super features, i think it's fine to remove

mparent61:
Since we already have the "exclude" query param logic, why not keep it, but have the view always append SF doc type to the exclude list? Seems like `exclude` would be handy in the future.

MichaelButkovic:
sure, that's fair

kand:
👍

mparent61:
👍

PR #332, opened 2016-09-12
Special coverage Video player preamble change

MelizzaP:
video player on special coverage landing page should have preamble "Sponsored by"

test link: http://onion-sponsored-by.test.theonion.com/special/boldness should have the campaign display preamble "Sponsored by"


bulbs/content/templatetags/content.py
```diff
@@ -38,3 +38,13 @@ def content_tunic_campaign_url(campaign_id,
 @register.simple_tag(takes_context=True)
 def build_video_share_uri(context, video_id):
     return context["request"].build_absolute_uri("/v/" + format(video_id))
+
+
+@register.simple_tag
+def campaign_display_preamble(tunic_campaign_id):
+    tunic_campaign_whitelist = [563]
```
mparent61: Not a good idea to hardcode this list, especially a library.

We should brainstorm how to avoid this. What about a new Tunic Campaign field set via Tunic CMS, and passed to client?

MelizzaP: I was afraid you'd say that

kand:
👍

mparent61:
👍

spra85:
👍

PR #335, opened 2016-09-13
updated the entries HTML endpoint

collin:
@mparent61 

* filters entries by published
* takes an array of entry ids rather than a timestamp


collin:
Will merge this into the main PR, just wanted an isolated review.

bulbs/liveblog/views.py
```diff
@@ -1,23 +1,26 @@
...
 
 
-class LiveblogNewEntriesView(TemplateView):
+class LiveblogNewEntriesView(BaseContentDetailView):
+    model = LiveBlogModel
     template_name = 'liveblog/new_entries.html'
+    redirect_correct_path = False
 
-    def get_context_data(self, slug, pk):
+    def get_context_data(self, object):
```
mparent61: should you call parent (super) class method first? (Do you want any of parent's context values?)

bulbs/liveblog/views.py
```diff
@@ -1,23 +1,26 @@
...
-        when = dateparse.parse_datetime(self.request.GET['new_as_of'])
-        if not when:
-            raise ValueError('param new_as_of must be an ISO formatted timestamp')
-        context['entries'] = liveblog.entries.filter(published__gte=when)
+        if 'entry_ids' not in self.request.GET:
+            raise ValueError('param "entry_ids" MUST be specified')
+        parsed_entry_ids = [x.strip() for x in self.request.GET['entry_ids'].split(',')]
+        context['entries'] = self.object.entries.filter(
+            pk__in=parsed_entry_ids,
+            published__lte=timezone.now())
```
mparent61: Does this also filter when "published=None" ? (my None comparison logic needs refresher)

bulbs/liveblog/views.py
collin: No, don't need any of the other data, special coverage stuff etc.

mparent61:
LGTM

bulbs/liveblog/views.py
collin: @mparent61 added a test that confirms `published=None` are filtered out.

PR #336, opened 2016-09-13
add data-track-action to share tools, needed for click tracking

MelizzaP:
Click tracking isn't working on the homepage player, because `data-track-action` is required.


collin:
:+1:

